PR #1, opened 2018-04-23
Initial setup

sbesson:
### Description

This PR proposes the migration of the CLI metadata functionality into a separate decoupling plugin. The goal here is to facilitate the development and release lifecycle of this plugin which currently needs to be rebased on the different development lines.

1cc4fb7ef181d0961b75d150616a3a9066c3e76a is the result of the filtering of the main metadata population classes from the current HEAD of `develop` (https://github.com/openmicroscopy/openmicroscopy/commit/d53b236345bc3fdc3f9047aa19874ae499a4a1e0) in order to consider

### Changes

This PR adds the minimal infrastructure:

- to have this component installed and deployed on PyPI following the [Python packaging principle](https://packaging.python.org/distributing/). Most of the classes (`setup.py`) were copied from https://github.com/ome/omero-cli-render
- to have this CLI plugin tested by https://github.com/openmicroscopy/omero-test-infra/ including the migration of the classes to the top-level package to prevent shadowing.

With this change, Travis should test the package installation as well as the CLI plugin integration tests. Manual tests can be performed on the OMERO 5.4 line by running 

```
bin/omero metadata -h  # should fail assuming OMERO_DEV_PLUGINS is not set
pip install https://github.com/sbesson/omero-metadata/archive/travis_setup.zip
bin/omero metadata -h
...
```

### Known limitations/questions

- only the CLI tests are included in this first round. Include the remaining `tabletests` and `metadatatest`  for `0.1.0`?
- `populate_metadata`: given the renaming in the root package, would `omero_populate_metadata` be more suitable?
- `ThreadPool` is currently imported from `populate_roi` but this dependency is largely unnecessary. Should it be migrated to an upstream `omero.util` class and/or copied here for simplicity?
- once released, timeline for removing `metadata.py` from the mainline to prevent amibuguity when `OMERO_DEV_PLUGINS` is set and this plugin is pip-installed

/cc @emilroz


sbesson:
I think there are 2 main possibilities:
- have this code only exposing CLI-level functionalities
- maintain a separation between a standalone utility API (`populate_metadata`) and the CLI wrappers (`omero_cli_metadata.py`)
Happy to gather thoughts about this restructuring beyond `0.1.0`.

jburel:
We also have the populate metadata script under ome/scripts

sbesson:
Pushed. Re https://github.com/ome/omero-cli-metadata/pull/1#issuecomment-383630928, looking at all public occurences on [GitHub search](https://github.com/search?q=omero+populate_metadata&type=Code) and more generally Google, there are definitely multiple locations consuming the `omero.util.populate_metadata` API.

I think four classes of options:

1. keep this PR as it i.e. migrate both the core metadata population and the CLI implementation in this new component. All other components  would have to declare `omero(-cli?)-metadata` as dependency via a `requirements.txt` file or similar. Note in the case of OMERO.scripts, this might get us into reviewing the deployment of scripts and the submodule dependency.
2. migrate the core metadata population classes to a new layer e.g.
`OMERO.py -> omero_api_metadata -> omero-cli-metadata, scripts, mapr`. For consumption, the new components would need to be declared as a dependency with similar considerations as in option 1.
3. leave `omero.util.populate_metadata` in OMERO.py and have it consumed by various downstream components (CLI, scripts, mapr...). This has limited deployment impact since CLI plugins, scripts and apps all have an implicit dependency on OMERO.py. Main downside of this option is the loss of the flexibility of options 1 and 2 as any improvement to the core API like https://github.com/openmicroscopy/openmicroscopy/pull/5691 would need to be scheduled as part of a full OMERO release. 
4. if option 3 is preferable for the core population API and decoupling the CLI implementation alone has little value and we could consider tabling this investigation.

sbesson:
Having discussed briefly the next steps for this repository especially with regard to the upcoming Dundee 2018 workshop preparation, below is a proposal for a series of immediate and future milestones:

| Feature                                                              | Commits/PRs                                                                               | Proposed tag        |
|----------------------------------------------------------------------|-------------------------------------------------------------------------------------------|---------------------|
| Filter the OMERO metadata53 branch                                   | https://github.com/sbesson/omero-cli-metadata/commit/15d3ddf59b949d95d25eab12461e9023cb44d91f                         | v0.1.0              |
| Filter the OMERO develop/metadata54 branch                           | https://github.com/ome/omero-cli-metadata/commit/1cc4fb7ef181d0961b75d150616a3a9066c3e76a | v0.2.0              |
| Add Python (setup.py) and OMERO testing infrastructure (.travis.yml) | https://github.com/ome/omero-cli-metadata/pull/1                                          | v0.2.1              |
| Migrate csv resource files in this repository                        |                                                                                           | patch release       |
| Activate testing of non-CLI tests (tables, metadata)                 |                                                                                           | patch release       |
| Migrate OMERO populate_metadata script                               |                                                                                           | patch/minor release |
| Add support for P/D/I bulk annotations                               |                                                                                           | minor release       |
| Port child options from v0.1.0?                                      |                                                                                           |                     |

Tags `0.1.0` and `0.2.0` only capture the state of the non-decoupled code for the main upstream branches. `0.2.1` is the first version that can be deployed to pypi.org and consumed as an external CLI plugin on OMERO 5.4. The following tasks are listed in no particular order and can be prioritized as a function of the deadlines/drivers.

manics:
We've previously discussed the badly needed refactoring of `populate_metadata.py`. How about leave `populate_metadata.py` in OMERO.py and instead add the refactored version to omero-cli-metadata (or omero-api-metadata)? Once it's fully functional we can deprecate `OMERO.py`'s `populate_metadata.py`

sbesson:
@manics agreed, I am not intending to modify upstream `omero.util.populate_metadata` which is consumed in many places and will need to follow the usual deprecation route.

The logic has been copied here - see `src/populate_metadata.py` and I would expect the refactoring will need to be included in the roadmap as well (adding to the table above).

dominikl:
Looks good. Tested it on idr-testing to annotate the HPA (idr0043) dataset 👍

PR #5, opened 2018-08-20
Bulk annotations and streaming

kkoz:
Added changes to support bulk annotations for projects and datasets, and to change csv reading to process the data line by line instead of buffering the data in memory and writing to the table all at once.

### DIFF
--- a/src/omero_cli_metadata.py
+++ b/src/omero_cli_metadata.py
@@ -476,20 +476,20 @@ class MetadataControl(BaseControl):
                 cfgid = cfgann.getFile().getId()
                 md.linkAnnotation(cfgann)
 
+        loops = 0
+        ms = 0
+        if not args.dry_run:
+            wait = args.wait
+            if wait:
+                ms = 5000
+                loops = int((wait * 1000) / ms) + 1
+
         # Note some contexts only support a subset of these args
         ctx = context_class(client, args.obj, file=args.file, fileid=fileid,
                             cfg=args.cfg, cfgid=cfgid, attach=args.attach,
-                            options=localcfg)
+                            options=localcfg, batch_size=args.batch,
+                            loops=loops, ms=ms)
         ctx.parse()
-        if not args.dry_run:
-            wait = args.wait
-            if not wait:
-                loops = 0
-                ms = 0
-            else:
-                ms = 5000
-                loops = int((wait * 1000) / ms) + 1
-            ctx.write_to_omero(batch_size=args.batch, loops=loops, ms=ms)
 
     def rois(self, args):
         "Manage ROIs"
--- a/src/populate_metadata.py
+++ b/src/populate_metadata.py
@@ -1921,7 +1921,6 @@ if __name__ == "__main__":
     password = None
     hostname = 'localhost'
     port = 4064  # SSL
-    info = False
     session_key = None
     logging_level = logging.INFO
     thread_count = 1
@@ -1936,8 +1935,6 @@ if __name__ == "__main__":
             hostname = argument
         if option == "-p":
             port = int(argument)
-        if option == "-i":
-            info = True
         if option == "-k":
             session_key = argument
         if option == "-d":
--- a/src/populate_metadata.py
+++ b/src/populate_metadata.py
@@ -880,8 +880,8 @@ class ParsingContext(object):
         self.column_types = column_types
         self.value_resolver = ValueResolver(client, target_object)
         self.parsing_util_factory = ParsingUtilFactory(client,
-                                                  target_object,
-                                                  self.value_resolver)
+                                                       target_object,
+                                                       self.value_resolver)
 
     def create_annotation_link(self):
         self.target_class = self.target_object.__class__
--- a/src/populate_metadata.py
+++ b/src/populate_metadata.py
@@ -99,6 +99,13 @@ WELL_NAME_COLUMN = 'Well Name'
 DATASET_NAME_COLUMN = 'Dataset Name'
 IMAGE_NAME_COLUMN = 'Image Name'
 
+ADDED_COLUMN_NAMES = [PLATE_NAME_COLUMN,
+                      WELL_NAME_COLUMN,
+                      DATASET_NAME_COLUMN,
+                      IMAGE_NAME_COLUMN,
+                      'image']
+
+
 COLUMN_TYPES = {
     'plate': PlateColumn, 'well': WellColumn, 'image': ImageColumn,
     'dataset': DatasetColumn, 'roi': RoiColumn,
@@ -106,6 +113,9 @@ COLUMN_TYPES = {
 }
 
 REGEX_HEADER_SPECIFIER = r'# header '
+
+DEFAULT_TABLE_NAME = 'bulk_annotations'
+MAX_COLUMN_COUNT = 512
 
 
 class Skip(object):
@@ -821,10 +831,34 @@ class ProjectWrapper(PDIWrapper):
         log.debug('Completed parsing project: %s' % self.target_object.id.val)
 
 
+class ParsingUtilFactory(object):
+
+    def get_filter_for_plate(self, column_index, target_name):
+        return lambda row: True if row[column_index] == target_name else False
+
+    def get_generic_filter(self):
+        return lambda row: True
+
+    def __init__(self, client, target_object, value_resolver):
+        self.target_object = target_object
+        self.target_class = target_object.__class__
+        self.value_resolver = value_resolver
+
+    def get_value_resolver(self):
+        return self.value_resolver
+
+    def get_filter_function(self, column_index=-1):
+        if PlateI is self.target_class and column_index != -1:
+            return self.get_filter_for_plate(
+                column_index, unwrap(self.target_object.getName()))
+        else:
+            return self.get_generic_filter()
+
+
 class ParsingContext(object):
     """Generic parsing context for CSV files."""
 
-    def __init__(self, client, target_object, file=None, fileid=None,
+    def __init__(self, client, target_object, parsing_util_factory, file=None, fileid=None,
                  cfg=None, cfgid=None, attach=False, column_types=None,
                  options=None):
         '''
@@ -844,7 +878,8 @@ class ParsingContext(object):
         self.target_object = target_object
         self.file = file
         self.column_types = column_types
-        self.value_resolver = ValueResolver(self.client, self.target_object)
+        self.parsing_util_factory = parsing_util_factory
+        self.value_resolver = self.parsing_util_factory.get_value_resolver()
 
     def create_annotation_link(self):
         self.target_class = self.target_object.__class__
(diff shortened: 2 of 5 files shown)


test/integration/tablestest/test_populate_metadata.py
```diff
@@ -24,12 +24,16 @@
 
 """
 
+import csv
```
sbesson: This import causes the Python code compliance step to fail in Travis

sbesson:
@kkoz thanks for your first contribution to this plugin. I commented on one import which causes the Travis build to fail. Once CI passes, we can look into testing these changes functionally. A couple of immediate questions:

- I assume this change should be tested against examples of P/D as well as S/P/W for non-regression. We have certainly examples of bulk annotations for both types of hierarchies in IDR
- briefly scanning through the changes, the API changes in `populate_metadata` probably mean this would be released as `0.3.0`
- it looks like `info` argument is now largely a no-op as `write_to_omero` is no longer conditional. From the perspective of the CLI plugin this is entirely fine as this argument was never exposed.

kkoz:
@sbesson Thanks for the notes. I will be making a commit shortly with the unnecessary import removed. For future reference, is there any way to check if my code will pass a travis-ci check without doing a commit? I've responded to your questions/comments below.

-  The code should be tested against P/D and S/P/W data.

-  That's fine - I will update CHANGES.rst

- Removed info from command line args. Please let me know if there's anything else that needs to be done on this front.

test/integration/tablestest/test_populate_metadata.py
```diff
@@ -29,7 +29,9 @@
 
 from omero.model import PlateI, WellI, WellSampleI, OriginalFileI
 from omero.rtypes import rint, rstring, unwrap
-from omero.util.populate_metadata import ParsingContext
+from omero.util.populate_metadata import (ParsingContext,
```
sbesson: Just realizing that this still imports the legacy OMERO.py classes rather than the class in the package. Largely, something I missed in #1 as I did not activate these tests. Can you try to update this line to

```
from populate_metadata import (ParsingContext,
...
``

and similarly in `test/integration/metadata/test_populate.py` so that your new factory class is picked up?

NB: in a follow-up PR, we'll probably want to look into properly namespacing these packages.

**Comment by sbesson** [comment]
@kkoz for pre-testing, assuming you have Docker installed, you should be able to test your changes locally by running the same calls as the ones executed in `.travis.yml`:

```
$ git clone git://github.com/openmicroscopy/omero-test-infra .omero
$ .omero/cli-docker
```

This is going to consume the omero-test-infra infrastructure to 1- create a Docker-based OMERO environment  (DB, server, web), 2- install the plugin and run Python compliance checks, 3- execute the integration tests.

**Comment by kkoz** [comment]
@sbesson Thanks for the info. I ran `.omero/cli-docke` but the output is mostly flake8 notes like `./omero-metadata/venv/lib/python2.7/distutils/__init__.py:100:1: E305 expected 2 blank lines after class or function definition, found 0`
There are a lot of these: `flake8.main.application   MainProcess 130168 INFO     Found a total of 108169 violations and reported 104755`
Is there some way for me to disable the flake8 checks so I can see the output of the tests?
Also, which version of populate_metadata.py is being used in the container? How can I make sure it's the one I want to test? I'll perform another commit with the import changes in the meantime. Thanks again for your help.

**Comment by sbesson** [comment]
@kkoz looks great now. Re `flake8`, you can disable it by updating `.omero/py-check` and commenting the `flake8` line. Is the `venv` folder your local development virtual environment? If so, it should be possible to add a file excluding the check of known folders.

The version of `populate_metadata.py` tested should be the local version. The local directory is copied in the docker container via `docker cp` and installed there via `python setup.py sdist` before executing the tests.

File: src/populate_metadata.py [hunk 3]
```diff
@@ -751,15 +824,42 @@ def _load(self):
...
+        else:
+            return self.get_generic_filter()
+
+
 class ParsingContext(object):
     """Generic parsing context for CSV files."""
 
-    def __init__(self, client, target_object, file=None, fileid=None,
+    def __init__(self, client, target_object,
+                 parsing_util_factory, file=None, fileid=None,
```
sbesson: Is `parsing_util_factory` a new mandatory argument in `ParsingContext` as proposed here or an optional key/value pair as suggested in the tests. Trying to consume this branch against a Project from the CLI wrapper:

```
bash-4.2$ pip install git+git://github.com/kkoz/omero-metadata.git@bulk-annotations-and-streaming#egg=omero-metadata --user
Collecting omero-metadata from git+git://github.com/kkoz/omero-metadata.git@bulk-annotations-and-streaming#egg=omero-metadata
...
bash-4.2$  OMERO-server/OMERO.server/bin/omero metadata populate --report --batch=10000 --file IDR-import/idr0021-lawo-pericentriolarmaterial/experimentA/idr0021-experimentA-annotation.csv Project:1
Using session for root@localhost:4064. Idle timeout: 10 min. Current group: system
  File "OMERO-server/OMERO.server/bin/omero", line 125, in <module>
    rv = omero.cli.argv()
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1095, in invoke
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1172, in onecmd
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "/home/omero/.local/lib/python2.7/site-packages/omero_cli_metadata.py", line 482, in populate
    options=localcfg)
TypeError: __init__() takes at least 4 arguments (9 given)
```

NB: this also reveals the absence of integration test for the CLI `metadata populate` subcommand which would have picked up this error. This is independent of this PR but I will try and produce an extra commit to cover this use case.

kkoz: The intention was to have it be mandatory. This was probably a bit of premature refactoring on my part. I will see if I can change the code to restore the signature of `__init__` in ParsingContext. My eventual goal was to refactor the code such that it allowed for unit testing. But that is outside the scope of this PR so I'll try to change it.

sbesson:
A couple of notes from the discussion with @emilroz @kkoz and @joshmoore. Generally, in the absence of any other outstanding requirement, we are starting to define the roadmap for a `0.3.0` release of `omero-metadata` including:

- the support of large CSV (multiple 10-100G)
- better support/fixes of the behavior for various column types
- additional API cleanup/refactoring including =the usage of the new factory class
- more unit and integration tests (#8)

In terms of functional testing perspective, while the latest `omero-metadata:0.2.2` can keep been used in production, different types of IDR studies e.g. [idr0021](https://github.com/IDR/idr0021-lawo-pericentriolarmaterial) and [idr0008](https://github.com/IDR/idr0008-rohn-actinome) should be used to confirm the various population contexts (bulk annotation population, map annotation population, map annotation deletion) remain functional.

Testing the last commit, the default `metadata populate` command failed as the CLI wrapper is expecting `ctx.write_to_omero` which is removed by this PR:

```
bin/omero metadata populate --report --batch=1000 --file=../../IDR-import/screenA/idr0037-screenA-annotation.csv Screen:52
Using session for root@localhost:4064. Idle timeout: 10 min. Current group: system
DEBUG:omero.util.populate_metadata:Loading Screen:52
...
  File "bin/omero", line 125, in <module>
    rv = omero.cli.argv()
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1095, in invoke
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1172, in onecmd
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "/home/omero/.local/lib/python2.7/site-packages/omero_cli_metadata.py", line 492, in populate
    ctx.write_to_omero(batch_size=args.batch, loops=loops, ms=ms)
```

As discussed, no major objection to removing the `write_to_omero` API as long as the CLI options (`--batch`, `--loops`, `--ms`) remain functional. The simplest option might be to add these to the context constructor and handle them internally depending on the type of operation.

kkoz:
@sbesson I made changes to solve the problem, but I'm having trouble testing them. When I run:
`bin/omero metadata populate --report --batch=1000 --file=/home/kevin/testdata/test.csv Plate:1`
I get:
`bin/omero: error: argument <subcommand>: invalid choice: 'metadata'`
Also it might be good to have an IDR csv file to test with. How can I get one?

sbesson:
@kkoz about the invalid choice issue, immediate question: does `pip freeze` list you the `omero-metadata` module?
All of the IDR CSV files are located in https://github.com/IDR/idr-metadata and its submodules - we are moving towards 1 repository per IDR study. For testing, two "small" studies we have been using are:

- https://github.com/IDR/idr0021-lawo-pericentriolarmaterial/blob/master/experimentA/idr0021-experimentA-annotation.csv for an example of P/D/I bulk annotation
- https://github.com/IDR/idr0008-rohn-actinome/blob/master/screenB/idr0008-screenB-annotation.csv for an example of HCS bulk annotation

For both of these, the raw data is also available if needed.

kkoz:
@sbesson I successfully imported the idr0021datasets into omero. I added them to a project I created and ran `./bin/omero metadata populate --report --batch=1000 --file=/home/kevin/code/idr0021-lawo-pericentriolarmaterial/experimentA/idr0021-experimentA-annotation.csv Project:201` without any issues. The project now has a "bulk_annotations" attachment but none of the datasets or images have Tables with data associated with them. Did I do use the populate metadata command incorrectly, or is this unexpected behavior? I am using OMERO.web 5.4.6-ice36-b87.

sbesson:
@kkoz excellent. I'll have a go at retesting the annotation workflow (might be tomorrow).
I think the absence of tabs in OMERO.web 5.4.6 is expected and this is what https://github.com/openmicroscopy/openmicroscopy/pull/5832 is trying to fix. I will try to use the latter to confirm via the Web UI that the metadata population worked as expected.

sbesson:
Executed a similar workflow to https://github.com/ome/omero-metadata/pull/5#issuecomment-417591581 against a screen (`idr0008/screenB`). Both table population and table to map annotation conversion worked without issue. I was not able to fully test the deletion operation due to time issue. However we know this is one for which can be improved.

I will perform a final round of code review on Monday.

src/omero_cli_metadata.py
```diff
@@ -476,20 +476,20 @@ def populate(self, args):
...
                 md.linkAnnotation(cfgann)
 
-        # Note some contexts only support a subset of these args
-        ctx = context_class(client, args.obj, file=args.file, fileid=fileid,
-                            cfg=args.cfg, cfgid=cfgid, attach=args.attach,
-                            options=localcfg)
-        ctx.parse()
+        loops = 0
+        ms = 0
         if not args.dry_run:
```
sbesson: If I understand correctly the `--dry-run` argument is now largely a no-op, similarly to the `-info` command. Primary question is whether we might want to resurrect it at some point via some constructor option of equivalent.

src/populate_metadata.py
```diff
@@ -99,14 +99,24 @@ def usage(error):
 DATASET_NAME_COLUMN = 'Dataset Name'
 IMAGE_NAME_COLUMN = 'Image Name'
 
+ADDED_COLUMN_NAMES = [PLATE_NAME_COLUMN,
+                      WELL_NAME_COLUMN,
+                      DATASET_NAME_COLUMN,
+                      IMAGE_NAME_COLUMN,
+                      'image']
```
sbesson: I'll update this line in my #10 once merged

src/populate_metadata.py
```diff
@@ -803,87 +905,194 @@ def get_column_widths(self):
...
+        table.close()
+
+    def create_file_annotation(self, table):
+        sf = self.client.getSession()
+        group = str(self.value_resolver.target_group)
+        update_service = sf.getUpdateService()
+
+        original_file = table.getOriginalFile()
+        file_annotation = FileAnnotationI()
+        file_annotation.ns = rstring(
```
sbesson: Unrelated to this PR, it should be possible to consume `omero.constants.namespaces.NSBULKANNOTATIONS`

PR #11, opened 2018-09-10
Implement dry-run option in delete map context class

sbesson:
As a preamble of upcoming work on the performance of the map annotation delete command, this PR:

- unifies the constructor of all the `Context` classes to support a `dry_run` boolean keyword, set to `False` by default
- updates the `DeleteMapAnnotationContext` OMERO deletion method to consume the `dry_run` argument
- updates the CLI `metadata populate` command to propagate the `dry-run` argument to the context class

This should allow to use `bin/omero metadata populate --context deletemap --dry-run` to measure a baseline and compare performance metrics for variations of the deletion commands without deleting the actual annotations on the server.

17e3737 updates the integration tests for the map deletion context class to test `ctx.parse()` with and without `--dry-run`

### DIFF
--- a/src/omero_cli_metadata.py
+++ b/src/omero_cli_metadata.py
@@ -478,11 +478,10 @@ class MetadataControl(BaseControl):
 
         loops = 0
         ms = 0
-        if not args.dry_run:
-            wait = args.wait
-            if wait:
-                ms = 5000
-                loops = int((wait * 1000) / ms) + 1
+        wait = args.wait
+        if wait:
+            ms = 5000
+            loops = int((wait * 1000) / ms) + 1
 
         # Note some contexts only support a subset of these args
         ctx = context_class(client, args.obj, file=args.file, fileid=fileid,
--- a/src/omero_cli_metadata.py
+++ b/src/omero_cli_metadata.py
@@ -488,7 +488,7 @@ class MetadataControl(BaseControl):
         ctx = context_class(client, args.obj, file=args.file, fileid=fileid,
                             cfg=args.cfg, cfgid=cfgid, attach=args.attach,
                             options=localcfg, batch_size=args.batch,
-                            loops=loops, ms=ms)
+                            loops=loops, ms=ms, dry_run=args.dry_run)
         ctx.parse()
 
     def rois(self, args):
--- a/src/populate_metadata.py
+++ b/src/populate_metadata.py
@@ -1899,7 +1899,7 @@ class DeleteMapAnnotationContext(_QueryContext):
             deleted = rsp.deletedObjects
             log.info("Deleted objects in %g s" % (time.time() - start))
             for k, v in deleted.iteritems():
-                log.info("  %d %s", k, len(v))
+                log.info("  %d %s", len(v), k)
                 log.debug("  %s %s", k, v)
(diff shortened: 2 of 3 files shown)


dominikl:
The code looks fine (as far as I can tell). But testing on test52 gives me a `LockTimeout` (immediately):
```
[dlindner@test52-omeroreadwrite experimentA]$ /opt/omero/server/OMERO.server/bin/omero metadata populate --context deletemap --cfg idr0042-experimentA-bulkmap-config.yml --dry-run Project:402
INFO:omero.util.populate_metadata:Total MapAnnotationLinks in ['openmicroscopy.org/omero/bulk_annotations', 'openmicroscopy.org/omero/bulk_annotations/config', 'openmicroscopy.org/mapr/organism']: 4598
  File "/opt/omero/server/OMERO.server/bin/omero", line 130, in <module>
    rv = omero.cli.argv()
  File "/opt/omero/server/OMERO.server/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "/opt/omero/server/OMERO.server/lib/python/omero/cli.py", line 1095, in invoke
  File "/opt/omero/server/OMERO.server/lib/python/omero/cli.py", line 1172, in onecmd
  File "/opt/omero/server/OMERO.server/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "/home/dlindner/omero-metadata/src/omero_cli_metadata.py", line 491, in populate
    ctx.parse()
  File "/home/dlindner/omero-metadata/src/populate_metadata.py", line 1711, in parse
    self.write_to_omero()
  File "/home/dlindner/omero-metadata/src/populate_metadata.py", line 1878, in write_to_omero
    {"%sAnnotationLink" % objtype: batch}, self.loops, self.ms)
  File "/home/dlindner/omero-metadata/src/populate_metadata.py", line 1893, in _write_to_omero_batch
    del_cmd, loops=loops, ms=ms, failontimeout=True)
  File "/opt/omero/server/OMERO.server/lib/python/omero/clients.py", line 899, in submit
    closehandle=True)
  File "/opt/omero/server/OMERO.server/lib/python/omero/clients.py", line 918, in waitOnCmd
    callback.loop(loops, ms)  # Throw LockTimeout
  File "/opt/omero/server/OMERO.server/lib/python/omero/callbacks.py", line 256, in loop
    5000L, int(waited))
omero.LockTimeout: exception ::omero::LockTimeout
{
    serverStackTrace = None
    serverExceptionClass = None
    message = Command unfinished after 0.0 seconds
    backOff = 5000
    seconds = 0
}
```

sbesson:
The error above is known. This is because the default arguments for `populate --context deletemap` do not work. For now, you should be able to test by passing `--wait 100 --batch 100`.

I plan a follow-up PR to fix the default behavior although I could add a commit here if that's preferable.

PS: be careful when testing map annotation deletion against `test52`...

dominikl:
Tested against east-ci, works fine 👍

sbesson:
Thanks, as discussed earlier today, I'll probably open a follow-up PR to fix the default `--wait` time which might completely refactor the `deletemap` context into a `metadata delete` command.

PR #14, opened 2018-09-25
0.3.0a1 pre-release of omero-metadata

sbesson:
This PR proposes to release the 2 PRs currently merged in https://github.com/ome/omero-metadata/milestone/2?closed=1 and deploy it as a PyPI pre-release.

This is largely motivated by the fact working with map annotation deletion in the context of IDR now requires some of the changes included in #11 (to ignore `Annotation` during the deletion operation for primary keys)

Alternatively, a `0.3.0` release could be cut today and all future API breaking improvements moved to `0.4.0`.

/cc @dominikl @joshmoore @jburel @kkoz @emilroz

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ CHANGES
 0.3.0a1
 -------
 
+* Unify naming of the Image column in OMERO.tables
 * Fix CLI metadata populate --context deletemap --dry-run behavior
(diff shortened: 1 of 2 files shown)


dominikl:
Looks good. I'll test it with the HPA dataset after it's been released, if there are not objections 👍

PR #15, opened 2018-10-09
Only add the metadata for the specific dataset

dominikl:
With this PR only the metadata for the specific dataset is added to the bulkannotations table (if the target is a dataset). This will be needed for idr0043/hpa as there's going to be one big annotations.csv which will be extended continously as new datasets are added.

## Test
Create two datasets 'dataset1' and 'dataset2'

Import 'test1.fake' and 'test2.fake' into 'dataset1', and 'test3.fake' and 'test4.fake' into 'dataset2'.

Create an annotations.csv like this
```
Dataset Name,Image Name,Test
dataset1,test1.fake,test1
dataset1,test2.fake,test2
dataset2,test3.fake,test3
dataset2,test4.fake,test4
```

Run `./omero metadata populate --report --file annotations.csv Dataset:<id>` for both datasets.

Check that the bulkannotations table for dataset1 only contains the annotations for dataset1, same for dataset2.

Also check the workflow for Screen/Plate and make sure it works like before.

/cc @sbesson

### DIFF
--- a/src/populate_metadata.py
+++ b/src/populate_metadata.py
@@ -833,7 +833,7 @@ class ProjectWrapper(PDIWrapper):
 
 class ParsingUtilFactory(object):
 
-    def get_filter_for_plate(self, column_index, target_name):
+    def get_filter_for_target(self, column_index, target_name):
         return lambda row: True if row[column_index] == target_name else False
 
     def get_generic_filter(self):
@@ -847,10 +847,13 @@ class ParsingUtilFactory(object):
     def get_value_resolver(self):
         return self.value_resolver
 
-    def get_filter_function(self, column_index=-1):
+    def get_filter_function(self, column_index=-1, target_name=""):
         if PlateI is self.target_class and column_index != -1:
-            return self.get_filter_for_plate(
-                column_index, unwrap(self.target_object.getName()))
+            return self.get_filter_for_target(
+                column_index, target_name)
+        elif DatasetI is self.target_class and column_index != -1:
+            return self.get_filter_for_target(
+                column_index, target_name)
         else:
             return self.get_generic_filter()
 
@@ -942,14 +945,19 @@ class ParsingContext(object):
         if first_row_is_types:
             header_row = reader.next()
 
-        plate_header_index = -1
+        filter_header_index = -1
         for i, name in enumerate(header_row):
-            if name.lower() == 'plate':
-                plate_header_index = i
+            if PlateI is self.target_object.__class__ \
+                    and name.lower() == 'plate':
+                filter_header_index = i
+                break
+            if DatasetI is self.target_object.__class__ \
+                    and name.lower() == 'dataset name':
+                filter_header_index = i
                 break
 
         self.filter_function = self.parsing_util_factory.get_filter_function(
-            plate_header_index)
+            filter_header_index, self.value_resolver.wrapper.target_name)
 
         table = self.create_table()
         self.populate_from_reader(reader, self.filter_function, table, 1000)
@@ -1189,7 +1197,8 @@ class ParsingContext(object):
                     log.debug(image_column)
                     iname = image_name_column.values[i]
                     did = self.target_object.id.val
-                    if "dataset name" in columns_by_name:
+                    if "dataset name" in columns_by_name \
+                            and target_class is not DatasetI:
                         dname = columns_by_name["dataset name"].values[i]
                         did = self.value_resolver.wrapper.datasets_by_name[
                             dname].id.val


src/populate_metadata.py
```diff
@@ -1189,7 +1197,8 @@ def post_process(self):
                     log.debug(image_column)
                     iname = image_name_column.values[i]
                     did = self.target_object.id.val
-                    if "dataset name" in columns_by_name:
+                    if "dataset name" in columns_by_name \
+                            and target_class is not DatasetI:
                         dname = columns_by_name["dataset name"].values[i]
                         did = self.value_resolver.wrapper.datasets_by_name[
```
dominikl: I'm not sure what that's supposed to do, but calling it gives:
```
did = self.value_resolver.wrapper.datasets_by_name[
AttributeError: 'DatasetWrapper' object has no attribute 'datasets_by_name'
```
Therefore excluded that if the target is Dataset.

sbesson: This is resolving the target image i.e. currently it's trying to walk down the Project/Dataset/Image graph using:
- the project derived from the target passed as a command line argument
- then resolving the dataset ID from the dataset name column
- then resolving th image ID from the image name column

Here, your changes create a situation where you are passing the dataset both as the main target and as a column names.

sbesson:
Thanks for initating the discussion with a first proposal @dominikl. When it comes to partial (re)annotation, the current workflow involves deleting the table for the entire container and re-annotating completely. In the case of HPA, as we will be dealing with millions of images and thousands of datasets, this approach will probably hit scaling issues very quickly

Below, I tried to capture about the various implementation choices in terms of input/output/command for achieving a partial annotation workflow in the case of Projects/Datasets. These can be extrapolated to the case of HCS. Probably a first step is to decide which of these we will and will not support:

1. input layout for the annotation files:

    1. maintain a single global `annotation.csv(.gz)` for the top-level container
    1. maintain one `annotation.csv(.gz)` per annotation round e.g. one annotation csv for each incremental HPA import
    1. maintain one `annotation.csv` per intermediate container
   The last option is mentioned for completeness but is probably the less desirable as it will could create several thousands of files in the HPA case for instance.

1. OMERO annotations
    1. create the bulk annotation at the level of the top-level container e.g. the Project
    2. create the bulk annotation at the level of the intermediate containers e.g. Datasets
    Both approaches should be supported from the client perspective e.g. the Webclient will query tables at the Project and the Dataset level. In the first case, the challenge would be to appending to an existing table. In the second case, we would look into creating large amounts of smaller tables. In terms of querying, the second option might have some performance advantage although it creates more objects to manage.

1. partial annotation command semantics
  
    1. specify the subset to be annotated explicitly via the positional argument of the `populate` subcommand
      `bin/omero metadata populate --file annotation.csv.gz Dataset:<id>`
    2. specify the subset to be annotated explicitly via an extra command-line argument e.g. 
      `bin/omero metadata populate --file annotation.csv.gz --range <range> Project:<id>`
    3. specifying the subset to be annotated implicitly via the content of the annotation file. This would require 1iii i.e. create annotation.csv files only listing  the datasets to be annotated e.g.
     `bin/omero metadata populate --file annotation{_run1}.csv.gz Project:<id>`

The current proposal implements the following combination: 1i, 2ii and 3i. Immediate question is how the workflow would work in the case of 1000 Datasets. It feels like with this option, one would need to query the ID of all the datasets, then construct 1000 commands and call these commands in a loop.

dominikl:
1. input layout for the annotation files:
I think i and ii wouldn't make a difference for the plugin itself. I'd agree, iii is probably not a good choice. I'd actually prefer option ii, having one annotation.csv per import batch, that's more practical than having a multi-GB annotation.csv (which can be generated later if needed anyway).

2. OMERO annotations
Yes, as a first step I'd go for option ii only, bulk annotation table on the Dataset level. Afaik we don't have a case anywhere where we modifiy an existing table, don't know how feasible that would be.

3. partial annotation command semantics
I think the `bin/omero metadata populate --file annotation.csv.gz Dataset:<id>` should be clear enough. If there is an annotation.csv which has Dataset specific annotations and you specify a `Dataset:<id>` I think it's clear that you only want the annotations for this dataset. 
If we don't support 2. ii (extending the bulkannotation on the project level), then we wouldn't need a more complex syntax like `-file annotation.csv.gz --range <range> Project:<id>` either. 
This could be added later anyway, if really necessary.

Yes, after the bulkannotations are created, one has to call `bin/omero metadata populate --context bulkmap Dataset:<id>` with the specific Ids. I wonder if we could do something like `...  --context bulkmap Project:<id> --traverse` which then iterates through the datasets and only creates the missing map annotations. Does that get too hacky/messy?

sbesson:
Discussed with @dominikl today. This was tested as part of the annotation of a second set of 1000 HPA antibodies in `prod61`. This testing highlighted the current implementation has some performance downsides. Mostly, the same large `.cvs.gz` needs to be parsed and filtered as many times as there are datasets (1000) in that case, leading to unnecessary overhead.

For the third set of antibodies (currently scheduled to `prod63`) we will probably need to come back to this RFE and investigate a separate implementation to reduce the annotation time.

sbesson:
Tested using the two workflows: bulk annotation of datasets and bulk annotation of plates i.e.

```
bin/omero metadata populate Plate:<id> --file plates.tsv 
bin/omero metadata populate Dataset:<id> --file datasets.tsv 
```

Test using datasets and plates matching the TSV files passed with the datasets and plates individually annotated and the tables displayed at the images or wells level.

screen shot 2019-03-05 at 15 27 06
screen shot 2019-03-05 at 15 27 14

Tested also the annotation of projects and datasets in the case of mismatching information. When the target name is incorrect, the command error with a `MetadataError`. However when the target is correctly found but the images are mismatching, empty bulk annotations are created

screen shot 2019-03-05 at 15 27 23
screen shot 2019-03-05 at 15 27 27

The latter confirms a current limitation of the library as there is no reporting of mismatch between annotation file content and objects in the containers. This might be something we want to look into as a RFE. Immediately, merging this. I will open a package renaming PR similar to https://github.com/ome/omero-upload/pull/3 but then I propose to get the plugin released as `0.4.0` ?

sbesson:
Now released as https://pypi.org/project/omero-metadata/0.3.1/.

PR #16, opened 2018-11-22
Drop Python 2.6

jburel:
Drop python26

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,7 +13,7 @@ Requirements
 ============
 
 * OMERO 5.4.0 or newer
-* Python 2.7+
+* Python 2.7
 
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -13,7 +13,7 @@ Requirements
 ============
 
 * OMERO 5.4.0 or newer
-* Python 2.6+
+* Python 2.7+
```
manics: As mentioned on https://github.com/ome/omero-cli-duplicate/pull/6#pullrequestreview-177858246 could change to just `2.7` since `3` isn't supported

jburel:
@manics updated as suggested

PR #17, opened 2019-01-25
Allow multiple targets

dominikl:
Follow up to  #15 .

With this PR one could specifiy multiple targets (e. g. `./omero metadata populate --file test.csv Dataset:1,2,3,4,5`). For each target a bulk_annotations table will be created and attached, containing only the information specific to this target. This only requires one round of parsing through the provided csv file (well, I think it's actually two rounds, a pre-processing step and then the actual parsing).

/cc @sbesson

### DIFF
--- a/src/omero_cli_metadata.py
+++ b/src/omero_cli_metadata.py
@@ -369,21 +369,22 @@ class MetadataControl(BaseControl):
 
     def original(self, args):
         "Print the original metadata in ini format"
-        md = self._load(args)
-        try:
-            source, global_om, series_om = md.get_original()
-        except AttributeError:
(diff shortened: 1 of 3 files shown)


dominikl:
Ok, seems to break all integration tests, even the dataset ones :-/ Maybe not such a good idea then...

sbesson:
@dominikl does this PR effectively replace #15 or should both of them be reviewed together?

dominikl:
Will probably close it again. Sorry for the hassle...

dominikl:
Might give it another go... must be pretty close to get it working.

dominikl:
Closing. That's gonna turn into a complete rewrite of the plugin...

PR #18, opened 2019-02-27
Add roi utils for creating masks

manics:
Adds two helper methods for creating masks from binary and labelled image arrays.

### DIFF
--- a/src/omero_metadata/rois.py
+++ b/src/omero_metadata/rois.py
@@ -38,7 +38,7 @@ def mask_from_binary_image(
     """
     Create a mask shape from a binary image (background=0)
     :param numpy.array binim: Binary 2D array, must contain values [0, 1] only
-    :param rgba int-4-tuple: Optional (reg, green, blue, alpha) colour
+    :param rgba int-4-tuple: Optional (red, green, blue, alpha) colour
     :param z: Optional Z-index for the mask
     :param c: Optional C-index for the mask
     :param t: Optional T-index for the mask
@@ -101,7 +101,7 @@ def masks_from_label_image(
     """
     Create mask shapes from a label image (background=0)
     :param numpy.array labelim: 2D label array
-    :param rgba int-4-tuple: Optional (reg, green, blue, alpha) colour
+    :param rgba int-4-tuple: Optional (red, green, blue, alpha) colour
     :param z: Optional Z-index for the mask
     :param c: Optional C-index for the mask
     :param t: Optional T-index for the mask
--- a/src/omero_metadata/rois.py
+++ b/src/omero_metadata/rois.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+   ROI utils
+
+   Copyright (C) 2019 University of Dundee. All rights reserved.
+   Use is subject to license terms supplied in LICENSE.txt
+"""
+import numpy as np
+from omero.gateway import ColorHolder
+from omero.model import MaskI
+from omero.rtypes import (
+    rdouble,
+    rint,
+    rstring,
+)
+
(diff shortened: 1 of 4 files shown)


src/omero_metadata/rois.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Create a mask shape from a binary image (background=0)
+    :param numpy.array binim: Binary 2D array, must contain values [0, 1] only
+    :param rgba int-4-tuple: Optional (reg, green, blue, alpha) colour
+    :param z: Optional Z-index for the mask
+    :param c: Optional C-index for the mask
+    :param t: Optional T-index for the mask
+    :param text: Optional text for the mask
+    :param raise_on_no_mask: If True (default) throw an exception if no mask
+           found, otherwise return an empty Mask
```
manics: Not sure what the default should be for `raise_on_no_mask`.

src/omero_metadata/rois.py
```diff
@@ -0,0 +1,118 @@
...
+        super(Exception, self).__init__(msg)
+
+
+def mask_from_binary_image(
+        binim, rgba=None, z=None, c=None, t=None, text=None,
+        raise_on_no_mask=True):
+    """
+    Create a mask shape from a binary image (background=0)
+    :param numpy.array binim: Binary 2D array, must contain values [0, 1] only
+    :param rgba int-4-tuple: Optional (reg, green, blue, alpha) colour
```
mtbc: red

will-moore:
This sounds really useful (I always struggle to create masks for testing) but any pointers on how to use it would be great (without me having to read all the code)?
Also, to test a particular branch what do I need to do to enable this plugin? Build? Put on PYTHONPATH etc?

manics:
@will-moore I documented usage in the python method docs, but as you've pointed out that's displayed anywhere accessible. @sbesson Reckon we could experiment with generating docs on https://readthedocs.org/ ?

For testing: `pip install https://github.com/ome/omero-metadata/archive/22ccc9d46fdb34e87d7d96b46b84276c10b7f243.zip` or `pip install https://github.com/manics/omero-metadata/archive/roi-utils.zip`, adding to `PYTHONPATH` should also work.

sbesson:
Same thought as @will-moore, thanks for adding such utilities especially as we re-use the same pieces of code. 

Main initial feedback is that the `omero-metadata` was specifically designed to deal with annotations (tables and maps primarily). Adding ROI functionality feels a bit like shoehorning features in an existing project.

Several discussion had been started on a library/plugin to deal with ROIs - see https://github.com/openmicroscopy/design/pull/80 or https://github.com/IDR/idr-metadata/pull/292. Should we start opening this work in a separate dedicated project e.g. `omero-roi`? Is there code elsewhere that we would like to integrate? /cc @dom

Readthedocs is an interesting idea to use external infrastructure to host Python docs. Requires gaining knowledge so no objection to start experimenting on a new component.

manics:
If we're going for another repo we need to decide whether we want loads of small libraries or something like `omero-gateway-extras`.

src/omero_metadata/rois.py
sbesson: If I read correctly the code, no mask means no non-zero pixels found. Depending on the use case, I can certainly see two behaviors:

- return an empty Mask
- return `None` or throw an exception 

The former call is more friendly to the caller of this API as it ensures that a mask is always returned. However from an IDR consumer perspective, I am not sure that there is value in storing an empty Mask rather than no mask. Given this API already has some logic to find bounding boxes, I think raising exception by default but having the possiblity to change the behavior is totally acceptable.

manics:
This was moved to https://github.com/ome/omero-rois/tree/v0.2.0

PR #20, opened 2019-04-25
Adjust integration test for change in global metadata.

mtbc:
Adjusts integration test to no longer expect `OriginalMetadataRequest` to return `header: GlobalMetadata` since https://github.com/openmicroscopy/bioformats/pull/3332. Corresponds with https://github.com/openmicroscopy/openmicroscopy/pull/6015.


mtbc:
(When this PR's time is ripe, rerun Travis then it should pass.)

sbesson:
Now green. Thanks @mtbc

PR #28, opened 2019-07-10
Usage instructions in readme

will-moore:
Added usage instructions to the README, particularly for ```populate``` command.

See it staged at https://github.com/will-moore/omero-metadata/tree/usage_instructions_in_README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,106 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
 `OMERO_DEV_PLUGINS` environment variable should not be set to prevent
 conflicts when importing the Python module.
 
+Usage
+=====
+
+The plugin is called from the command-line using the `omero` command::
+
+    $ bin/omero metadata &lt;subcommand&gt;
+
+Help for each command can be shown using the ``-h`` flag.
+Objects can be specified as arguments in the format ```Class:ID```, such
+as ```Project:123```.
+
+Available subcommands are:
+
+ - ``allanns``: Provide a list of all annotations linked to the given object
+ - ``bulkanns``: Provide a list of the NSBULKANNOTATION tables linked to the given object
+ - ``mapanns``: Provide a list of all MapAnnotations linked to the given object
+ - ``measures``: Provide a list of the NSMEASUREMENT tables linked to the given object
+ - ``original``: Print the original metadata in ini format
+ - ``pixelsize``: Set physical pixel size
+ - ``populate``: Add metadata (bulk-annotations) to an object (see below)
+ - ``populateroi``: Add ROIs to an object
+ - ``rois``: Manage ROIs
+ - ``summary``: Provide a general summary of available metadata
+ - ``testtables``: Tests whether tables can be created and initialized
+
+populate
+--------
+
+This command creates an ``OMERO.table`` (bulk annotation) from a ``CSV`` file and links 
+the table as a ``File Annotation`` to a parent container such as Screen, Plate, Project
+or Dataset. It also attempts to convert Image or Well names from the ``CSV`` into
+Image or Well IDs in the ``OMERO.table``.
+
+The ```CSV``` file can be provided as local file with ``--file path/to/file.csv``
+or as an OriginalFile in OMERO with ``--fileid 123``.
+
+If you wish to ensure that ``number`` columns are created for numerical data, this will
+allow you to make numerical queries on the table.
+Column Types are::
+
+    'plate': PlateColumn, 'well': WellColumn, 'image': ImageColumn,
+    'dataset': DatasetColumn, 'roi': RoiColumn, 'd': DoubleColumn, 'l': LongColumn,
+    's': StringColumn, 'b': BoolColumn
+
+These can be specified in the first row of a ``CSV`` with a ``# header`` tag.
+
+NB: Column names should not contain whitespace if you want to be able to query
+by these columns.
+
+Examples:
+
+To add a table to a Screen, the ``CSV`` file needs to specify ``Plate Name`` and ``Well``::
+
+    $ bin/omero metadata populate Plate:1 path/to/screen.csv
+
+screen.csv::
+
+    # header well,plate,s,d,l,d
+    Well,Plate,Drug,Concentration,Cell_Count,Percent_Mitotic
+    A1,plate01,DMSO,10.1,10,25.4
+    A2,plate01,DMSO,0.1,1000,2.54
+    A3,plate01,DMSO,5.5,550,4
+    B1,plate01,Monastrol,12.3,50,44.43
+    B2,plate01,Drug3,1.1,1,15.4
+    B3,plate01,Water,20.1,44,12.3
+
+This will create an OMERO.table linked to the Screen like this::
+
+    Well	Plate	Drug	Concentration	Cell_Count	Percent_Mitotic	Well Name	Plate Name
+    9154	3855	DMSO	10.1	        10	        25.4	        a1	        plate01
+    9155	3855	DMSO	0.1	            1000	    2.54	        a2	        plate01
+    9156	3855	DMSO	5.5	            550	        4.0	            a3	        plate01
+    9157	3855	Monastrol	12.3	    50	        44.43	        b1	        plate01
+    9158	3855	Drug3	1.1	            1	        15.4	        b2	        plate01
+    9159	3855	Water	20.1	        44	        12.3	        b3	        plate01
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -32,6 +32,108 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
...
+===== ====== ====== ============== =========== ================ =========== ===========
+Well  Plate  Drug   Concentration  Cell_Count  Percent_Mitotic  Well Name   Plate Name
+===== ====== ====== ============== =========== ================ =========== ===========
+9154  3855   DMSO   10.1           10          25.4             a1          plate01
+9155  3855   DMSO   0.1            1000        2.54             a2          plate01
+9156  3855   DMSO   5.5            550         4.0              a3          plate01
+9157  3855   DrugX  12.3           50          44.43            b1          plate01
+===== ====== ====== ============== =========== ================ =========== ===========
+
+If the target is a Plate instead of a Screen, the ``Plate`` column is not needed.
```
manics: How about add something about the `Well` and `Plate` columns being automatically added with the discovered IDs, especially since this is different to the behaviour below where the Image name but not Image ID is used.

will-moore:
NB: as noted by @manics and discussed with @pwalczysko, the behaviour of ```Plate``` and ```Well``` columns is different from ```Dataset Name``` and ```Image Name``` as documented at https://github.com/ome/omero-metadata/issues/29

Just added more info about renaming of Well -> Well Name etc.

README.rst
```diff
@@ -32,6 +32,114 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
...
+- ``populate``: Add metadata (bulk-annotations) to an object (see below)
+- ``populateroi``: Add ROIs to an object
+- ``rois``: Manage ROIs
+- ``summary``: Provide a general summary of available metadata
+- ``testtables``: Tests whether tables can be created and initialized
+
+populate
+--------
+
+This command creates an ``OMERO.table`` (bulk annotation) from a ``CSV`` file and links
```
sbesson: It might be worth mentioning all the following only applies if the default context (`--context csv`) is is selected i.e. if the `ParsingContext` is used. If the bulkmap context (`--context bulkmap`) is selected, then the command creates map annotations from an OMERO.table.

will-moore: Any examples of config files for ```--context bulkmap```? I don't really want to mention this functionality if we can't show how to use it.

sbesson: This has been primarily driven for the use case of IDR so every experiment/screen in https://github.com/IDR/idr-metadata should contain a bulkmap configuration file.

That being said, I think it is fine for this PR to be limited to documenting the `csv` context.

will-moore: OK, I'll leave it as it is for this PR. Documenting everything else the ```metadata``` plugin does can be added as needed.

README.rst
```diff
@@ -32,6 +32,114 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
...
+- ``d``: ``DoubleColumn``, for floating point numbers
+- ``l``: ``LongColumn``, for integer numbers
+- ``s``: ``StringColumn``, for text
+- ``b``: ``BoolColumn``, for true/false
+- ``plate``, ``well``, ``image``, ``dataset``, ``roi`` to specify objects
+
+These can be specified in the first row of a ``CSV`` with a ``# header`` tag (see examples below).
+The ``# header`` row is optional. Default column type is ``String``.
+
+NB: Column names should not contain whitespace if you want to be able to query
```
pwalczysko: Is it a whitespace ? I think it is ``spaces``

README.rst
```diff
@@ -32,6 +32,114 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
...
+
+To add a table to a Project, the ``CSV`` file needs to specify ``Dataset Name``
+and ``Image Name``::
+
+    $ bin/omero metadata populate Project:1 path/to/project.csv
+
+project.csv::
+
+    # header s,s,d,l,s
+    Image Name,Dataset Name,Bounding_Box,Channel_Index,Channel_Name
```
pwalczysko: Bounding_Box -> Total_Area or ROI_Area

Bounding Box triggers too many questions like "What is it ?" - we should go for a moer intuitive parameter in this general info README

README.rst
```diff
@@ -32,6 +32,114 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
...
+    img-01.png,dataset01,0.0469,1,DAPI
+    img-02.png,dataset01,0.142,2,GFP
+    img-03.png,dataset01,0.093,3,TRITC
+    img-04.png,dataset01,0.429,4,Cy5
+
+This will create an OMERO.table linked to the Project like this:
+
+========== ============ ============ ============= ============ =====
+Image Name Dataset Name Bounding_Box Channel_Index Channel_Name Image
```
pwalczysko: Bounding_box - as above

pwalczysko:
Update the date in https://github.com/ome/omero-metadata/pull/28/files#diff-88b99bb28683bd5b7e3a204826ead112R152 ?

README.rst
```diff
@@ -32,6 +32,114 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
...
+    $ bin/omero metadata <subcommand>
+
+Help for each command can be shown using the ``-h`` flag.
+Objects can be specified as arguments in the format ``Class:ID``, such
+as ``Project:123``.
+
+Available subcommands are:
+
+- ``allanns``: Provide a list of all annotations linked to the given object
+- ``bulkanns``: Provide a list of the NSBULKANNOTATION tables linked to the given object
```
pwalczysko: Explain more explicitly the connection between the NSBULK... and OMERO.tables - maybe in a note below this table

README.rst
```diff
@@ -32,6 +32,114 @@ version 5.4.x of OMERO.py. When using the CLI metadata plugin, the
...
+- ``mapanns``: Provide a list of all MapAnnotations linked to the given object
+- ``measures``: Provide a list of the NSMEASUREMENT tables linked to the given object
+- ``original``: Print the original metadata in ini format
+- ``pixelsize``: Set physical pixel size
+- ``populate``: Add metadata (bulk-annotations) to an object (see below)
+- ``populateroi``: Add ROIs to an object
```
pwalczysko: Do we have an example of in what form does the ROI have to be offered in order to be populated ? @manics @sbesson  ?

sbesson: I think this feature is largely a legacy feature which came together with the decoupling of the `metadata` plugin:

- the ROI population API is still in upstream OMERO.py -see [omero.util.populate_roi](https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.1/components/tools/OmeroPy/src/omero/util/populate_roi.py)
- the `omero_metadata` plugin currently simply imports this API and exposes it as a subcommand
- in terms of functionality, the current version of `populate_roi` primarily supports ROIs generated by a few proprietary HCS file formats ([Flex](https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.1/components/tools/OmeroPy/src/omero/util/populate_roi.py#L353), [MIAS](https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.1/components/tools/OmeroPy/src/omero/util/populate_roi.py#L250), [InCell](https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.1/components/tools/OmeroPy/src/omero/util/populate_roi.py#L406))

Open for discussion but my feeling is we should rather keep this plugin focused its annotation functionalities (tables, maps, original metadata). As part of the IDR work, @manics started work on https://github.com/ome/omero-rois which might be a much more appropriate location for this particular API.

manics: How about deprecate `populateroi` (I've never figured out how it works or where it gets the ROIs from) and avoid mentioning it anywhere?

README.rst
manics: `Bounding_Box` implies a pair of `(x0,y0) (x1,y1)` coordinates giving the bounds of the box, this looks like a single number.

pwalczysko: @manics what @will-moore meant there is "the area encompassed by a bounding box" - that is why I think we should go for  variations of ``Area...`` in this example

README.rst
will-moore: I'll remove ```populateroi```. There's also the ```rois``` command. Not sure what that does either.

pwalczysko: @will-moore the rois command enables to delete rois in that container and list them in form ROI:ID. I think it is useful

pwalczysko:
reads fine now. Ready to merge FMPOV

PR #34, opened 2019-09-12
Do not warn on missing well/plate names when the target is non HCS

sbesson:
This PR attempts to reduce the verbosity of the `populate` command when annotating a project or a dataset with a CSV file.

To test it, set up a dataset with a few images a CSV file as described in https://github.com/ome/omero-metadata#populate and run the populate command:

```
omero metadata populate --file=test.csv Project:<id>
```

Without this PR the command should log  info-level statements about missing plate and well columns. With this PR included, these irrelevant statements should be gone.

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -1169,7 +1169,7 @@ class ParsingContext(object):
                     )
                 well_name_column.size = max(well_name_column.size, len(v))
                 well_name_column.values.append(v)
-            else:
+            elif (ScreenI is target_class or PlateI is target_class):
                 log.info('Missing well name column, skipping.')
 
             if image_name_column is not None and (
@@ -1242,7 +1242,7 @@ class ParsingContext(object):
                 v = self.value_resolver.get_plate_name_by_id(plate)
                 plate_name_column.size = max(plate_name_column.size, len(v))
                 plate_name_column.values.append(v)
-            else:
+            elif (ScreenI is target_class or PlateI is target_class):
                 log.info('Missing plate name column, skipping.')
 
 


will-moore:
Without this PR:
```
$ omero metadata populate Dataset:4919 --file /Users/willadmin/Downloads/fruit.csv 
Using session for user-3@merge-ci-devspace.openmicroscopy.org:4064. Idle timeout: 10 min. Current group: read-only-1
INFO:omero_metadata.populate:Created new table OriginalFile:378741
INFO:omero_metadata.populate:Missing well name column, skipping.
INFO:omero_metadata.populate:Missing plate name column, skipping.
... etc.
```
with this PR:
```
$ omero metadata populate Dataset:4919 --file /Users/willadmin/Downloads/fruit.csv 
Using session for user-3@merge-ci-devspace.openmicroscopy.org:4064. Idle timeout: 10 min. Current group: read-only-1
INFO:omero_metadata.populate:Created new table OriginalFile:378742
```
Good to merge.


NB: First I tried to use the ```fileid``` argument to pass the OriginalFile ID of the csv, but this seems to be ignored (I get the same if I omit the --fileId argument):
```
$ omero metadata populate Dataset:4919 --fileid 375122
Using session for user-3@merge-ci-devspace.openmicroscopy.org:4064. Idle timeout: 10 min. Current group: read-only-1
  File "/Users/willadmin/Virtual/omero/bin/omero", line 130, in <module>
    rv = omero.cli.argv()
  File "/Users/willadmin/Desktop/PYTHON/omero-py/src/omero/cli.py", line 1717, in argv
    cli.invoke(args[1:])
  File "/Users/willadmin/Desktop/PYTHON/omero-py/src/omero/cli.py", line 1170, in invoke
  File "/Users/willadmin/Desktop/PYTHON/omero-py/src/omero/cli.py", line 1247, in onecmd
  File "/Users/willadmin/Desktop/PYTHON/omero-py/src/omero/cli.py", line 1329, in execute
    args.func(args)
  File "/Users/willadmin/Virtual/omero/lib/python2.7/site-packages/omero_metadata/cli.py", line 491, in populate
    ctx.parse()
  File "/Users/willadmin/Virtual/omero/lib/python2.7/site-packages/omero_metadata/populate.py", line 988, in parse
    if self.file.endswith(".gz"):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

will-moore:
Looking at the code, it seems that only the ```--file``` argument is supported and if this is not set then you'll get the exception above. I'm not sure whether ```--fileid``` ever worked, but not now.
Probably best just to remove that usage from the documentation until it's fixed. I also want to add dev install instruction ```pip install .```...

PR #35, opened 2019-09-16
readme dev install

will-moore:
In testing https://github.com/ome/omero-metadata/pull/34#issuecomment-531734411
I found that the ```--fileid``` argument is not supported via the metadata CLI.
This removes it from the README, until it's fixed.

Also add dev install instructions, needed when developing or testing etc.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,8 +68,7 @@ the table as a ``File Annotation`` to a parent container such as Screen, Plate, 
 or Dataset. It also attempts to convert Image or Well names from the ``CSV`` into
 Image or Well IDs in the ``OMERO.table``.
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -143,6 +142,15 @@ Well  Plate  Drug   Concentration  Cell_Count  Percent_Mitotic  Well Name   Plat
 
 If the target is a Plate instead of a Screen, the ``Plate`` column is not needed.
 
+Developer install
+=================
+
+This plugin can be installed from the source code with::
+
+    $ cd omero-metadata
+    $ pip install .
```
manics: Does installing in editable mode `pip install -e .` work?

joshmoore: I'd _assume_ so.

PR #36, opened 2019-10-10
Python 3 support

manics:
This PR updates `omero-metadata` to be Python 3 compatible.

- run `futurize` on the code and fix various imports
- declare `omero-py`, `future`, `PyYAML` as dependencies and `mox3` as test dependency

Travis should test the installation of the plugin and as run the integration tests against the current Python 3 `omero-test-infra` environment.

Proposed tag: `0.5.dev1`

### DIFF
--- a/.omeroci/env
+++ b/.omeroci/env
@@ -1,0 +1,1 @@
+PLUGIN=metadata
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@
 language: python
 sudo: required
 
-env: PLUGIN=metadata
-
 services:
   - docker
 
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ Requirements
 ============
 
 * OMERO 5.4.0 or newer
-* Python 2.7
+* Python 2.7, 3.6
 
 
 Installing from PyPI
--- a/README.rst
+++ b/README.rst
@@ -68,8 +68,7 @@ the table as a ``File Annotation`` to a parent container such as Screen, Plate, 
 or Dataset. It also attempts to convert Image or Well names from the ``CSV`` into
 Image or Well IDs in the ``OMERO.table``.
 
-The ``CSV`` file can be provided as local file with ``--file path/to/file.csv``
-or as an OriginalFile in OMERO with ``--fileid 123``.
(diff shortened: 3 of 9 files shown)


joshmoore:
@manics : I re-ran travis since errors here are now showing up in unrelated calls to `bin/omero`. One failure I see is:

```
    def newTable(self, repoId, path, _ctx=None):
>       return _M_omero.grid.SharedResources._op_newTable.invoke(self, ((repoId, path), _ctx))
E       ValueError: context value must be a string
```

joshmoore:
Pushed a fix. Waiting on travis. If this goes green, shall we move it out of "draft" status?

manics:
Could do, though it's still not tested under Python 3.

joshmoore:
Fair enough, but even just having syntax compatible will help. Before merging, I'll look into whether `master` is used or if I'd need a pre-release tag anyway.

joshmoore:
Now we're up to ```    def addData(self, cols, _ctx=None):
>       return _M_omero.grid.Table._op_addData.invoke(self, ((cols, ), _ctx))
E       ValidationException: exception ::omero::ValidationException
E       {
E           serverStackTrace = 
E           serverExceptionClass = 
E           message = Maximum string length in column Well Type is 1
E       }``` which someone else was looking into today.

manics:
Feel free to continue pushing fixes!

dominikl:
Looks good 👍 

Installed with `pip install git+git://github.com/snoopycrimecop/omero-metadata@py3_ci#egg=omero-metadata`

Tested using merge-ci:
```
(test) dominik@nelson:~/test/bin$ ./omero -s merge-ci-devspace.openmicroscopy.org metadata summary Image:18188
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
Using session for user-3@merge-ci-devspace.openmicroscopy.org:4064. Idle timeout: 10 min. Current group: read-only-1
Image:18188
-----------
Name: Centrin_PCNT_Cep215_20110506_Fri-1545_0_SIR_PRJ.dv
Roi count: 24
Bulk annotations: 1
Measurement tables: 0
Parent: Dataset:3183
Source metadata: False
Global metadata: True
Series metadata: True
MapAnnotations: 6
```

PR #40, opened 2019-12-05
Improved support for annotations with Unicode characters

### LINKED ISSUE #39: Handle byte-order-mark (BOM) gracefully
If a CSV starts with a byte-order-mark `omero metadata populate` fails with an unhelpful message along the lines of
```
$ omero metadata populate --file idr0069-screenA-annotation.csv Screen:2251

...lib/python2.7/site-packages/omero_metadata/populate.py", line 926, in preprocess_from_handle
    % first_row[header_index])
Exception: Empty column header in CSV: Plate
```

It should either print out a helpful error message (BOM found) or just handle it.

sbesson:
This is primarily driven by the `idr0067` IDR submission as the [annotation file](https://github.com/IDR/idr0067-king-yeastmeiosis/blob/master/experimentA/idr0067-experimentA-annotation.csv contained Unicode characters for the genetic background. This PR generally improves support and test coverage for the creation of OMERO.tables from UTF-8 CSV files containing Unicode characters with and without BOM.

c2fb0f1 ensure the string is encoded to `utf-8` before calculating the maximum length of the string column
667ba18 enforces that CSV files are always opened with UTF-8 encoding and also uses context manager to close both file handles
44f69e9 also includes support for CSV files with a leading BOM character (Fixes #39)

The Unicode support was manually tested in an IDR annotation workflow from a Python 3 client environment with a Python 2 server environment. Integration tests have been added for both case (74da887 and  917251f ) and should pass in Travis CI.

A minimal testing workflow would be to import `test&plateCols=2.fake` which should create a plate with two wells and run `populate metadata Plate:<id> --file <file>.csv` with the two different [annotations.zip](https://github.com/ome/omero-metadata/files/3947048/annotations.zip).

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -994,17 +994,15 @@ class ParsingContext(object):
 
     def parse(self):
         if self.file.endswith(".gz"):
-            data_for_preprocessing = gzip.open(self.file, "rt")
-            data = gzip.open(self.file, "rt")
+            with gzip.open(self.file, 'rt', encoding='utf-8') as f1:
+                self.preprocess_from_handle(f1)
+                with gzip.open(self.file, 'rt', encoding='utf-8') as f2:
+                    return self.parse_from_handle_stream(f2)
         else:
-            data_for_preprocessing = open(self.file, 'rt')
-            data = open(self.file, 'rt')
-
-        try:
-            self.preprocess_from_handle(data_for_preprocessing)
-            return self.parse_from_handle_stream(data)
-        finally:
-            data.close()
+            with open(self.file, 'rt', encoding='utf-8') as f1:
+                self.preprocess_from_handle(f1)
+                with open(self.file, 'rt', encoding='utf-8') as f2:
+                    return self.parse_from_handle_stream(f2)
 
     def preprocess_data(self, reader):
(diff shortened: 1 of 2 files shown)


sbesson:
Relaunched https://travis-ci.org/ome/omero-metadata/builds/621146075 following the release of the OMERO 5.6.0-m4 Docker images including the OMERO.tables Unicode fix (https://github.com/ome/omero-py/pull/143). The `UnicodeDecodeError` error thrown by the test while reading the CS file is likely related to the default encoding (see also https://github.com/ome/omero-py/issues/142)   667ba18 fixes the tests by ensuring CSV files are always opened with a UTF-8 encoding.

dominikl:
Looks good, works as expected, thanks Seb 👍

sbesson:
Thanks @dom. If no objections, proposing to cut a development release of `omero-metadata 0.5.dev2` with this change included. /cc @manics @jburel @joshmoore

manics:
Sounds fine to me, any reason not to do a full release?

sbesson:
I think my major concern is the dependency of this package on a development release of `omero-py`. Once `omero-py 5.6.0` is released, I think all plugins can be scheduled for final release.

PR #46, opened 2020-08-21
Fix usage --report option in the omero-metadata CLI plugin

sbesson:
The behavior of the following subcommands is left unchanged:
- `rois`: pass the option to the graph command
- `bulkanns`, `measures`, `mapanns`, `allanns`: provide  additional information about the object

For the `populate` and `populateroi` command, `--report` was used as a way to increase the logging verbosity. This is now superseded by the more standard `-v/-q` which can be passed multiple times to control the logging verbosity e.g `omero metadata populate -qq`

The behavior of `omero metadata populate --report` is unchanged and equivalent to `omero metadata populate -v` but the option is removed from the usage in `omero metadata populate -h`

/cc @pwalczysko

### DIFF
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -9,6 +9,7 @@
 """
 from __future__ import division
 
+import omero_ext.argparse as argparse
 from builtins import str
 from builtins import range
 from builtins import object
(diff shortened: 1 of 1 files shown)


joshmoore (commented):
Code changes make sense to me. :+1: I imagine eventually we might want `-v/-q` on all the subcommands (and then there's question of whether or not `omero metadata (-v|-q)` should also be supported, but I assume those are outwith this PR.

sbesson:
In general, I am all for unifying the logger configuration across the board using the standard `-v[vv]/-q[qq]`. I can probably push another commit for the other subcommands as well.

PR #47, opened 2020-09-23
Add dependency on version 5.8 or more recent of omero-py

sbesson:
The primary motivation is to get rid of the DeprecationWarning thrown on every invocation of `omero metadata populate --context bulkmap` fixed in ome/omero-py@dcde6b2.

Also remove PyYAML and future dependencies pulled transitively - but see https://github.com/ome/omero-cli-render/pull/45#issuecomment-698170137

See also https://github.com/ome/omero-cli-render/pull/45

Proposed tag: `0.5.1`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -124,9 +124,7 @@ setup(
     keywords=['OMERO.CLI', 'plugin'],
     cmdclass={'test': PyTest},
     install_requires=[
-        'future',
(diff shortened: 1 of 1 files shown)


sbesson:
Closing as per the same rationale as https://github.com/ome/omero-cli-render/pull/45#issuecomment-713094838

PR #49, opened 2020-11-03
populate metadata at image-level

erindiel:
Adds an ImageWrapper to populate an OMERO.table at the image-level with ROI-level data. This requires an `roi` column containing ROI ID. Multiple rows per-ROI ID are allowed (see object-level case below), but ROIs with appropriate IDs must exist in image.

Note that the Tables dropdown is not populated at the image-level, but a `bulk annotations` file is added to attachments.

This can also be used to populate object-level analytics to refer to objects in a label image/mask associated with an ROI, where objects are encoded by their pixel value in the label image. There is currently no checking of this `object` column (for example, to confirm that these objects exist in the label image associated with the corresponding ROI), and this column need not exist at all. Once the label mask workflow is solidified, these checks might be helpful to implement.

An example has been added to README.

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -49,6 +49,7 @@ from omero.model import OriginalFileI, PlateI, PlateAnnotationLinkI, ScreenI
 from omero.model import PlateAcquisitionI, WellI, WellSampleI, ImageI
 from omero.model import ProjectAnnotationLinkI, ProjectI
 from omero.model import ScreenAnnotationLinkI
+from omero.model import ImageAnnotationLinkI
 from omero.model import MapAnnotationI, NamedValue
 from omero.grid import ImageColumn, LongColumn, PlateColumn, RoiColumn
 from omero.grid import StringColumn, WellColumn, DoubleColumn, BoolColumn
@@ -146,6 +147,10 @@ class HeaderResolver(object):
 
     DEFAULT_COLUMN_SIZE = 1
 
+    image_keys = {
+        'roi': RoiColumn
+    }
+    
     dataset_keys = {
         'image': ImageColumn,
         'image_name': StringColumn,
@@ -209,6 +214,9 @@ class HeaderResolver(object):
         elif ProjectI is target_class:
             log.debug('Creating columns for Project:%d' % target_id)
             return self.create_columns_project()
+        elif ImageI is target_class:
+            log.debug('Creating columns for Image:%d' % target_id)
+            return self.create_columns_image()
         raise MetadataError(
             'Unsupported target object class: %s' % target_class)
 
@@ -232,6 +240,9 @@ class HeaderResolver(object):
 
     def create_columns_project(self):
         return self._create_columns("project")
+
+    def create_columns_image(self):
+        return self._create_columns("image")
 
     def _create_columns(self, klass):
         if self.types is not None and len(self.types) != len(self.headers):
@@ -333,6 +344,8 @@ class ValueResolver(object):
             self.wrapper = ScreenWrapper(self)
         elif ProjectI is self.target_class:
             self.wrapper = ProjectWrapper(self)
+        elif ImageI is self.target_class:
+            self.wrapper = ImageWrapper(self)
         else:
             raise MetadataError(
                 'Unsupported target object class: %s' % self.target_class)
@@ -415,6 +428,8 @@ class ValueResolver(object):
         # Prepared to handle DatasetColumn
         if DatasetColumn is column_class:
             return self.wrapper.resolve_dataset(column, row, value)
+        if RoiColumn is column_class:
+            return self.wrapper.resolve_roi(column, row, value)
         if column_as_lower in ('row', 'column') \
            and column_class is LongColumn:
             try:
@@ -838,6 +853,58 @@ class ProjectWrapper(PDIWrapper):
         log.debug('Completed parsing project: %s' % self.target_object.id.val)
 
 
+class ImageWrapper(ValueWrapper):
+
+    def __init__(self, value_resolver):
+        super(ImageWrapper, self).__init__(value_resolver)
+        self.rois_by_id = dict()
+        self._load()
+    
+    def resolve_roi(self, column, row, value):
+        try:
+            return self.rois_by_id[int(value)].id.val
+        except KeyError:
+            log.warn('Image is missing ROI: %s' % value)
+            return Skip()
+        except ValueError:
+            log.warn('Wrong input type for ROI ID: %s' % value)
+            return Skip()
+
+    def _load(self):
+        query_service = self.client.getSession().getQueryService()
+        parameters = omero.sys.ParametersI()
+        parameters.addId(self.target_object.id.val)
+        log.debug('Loading Image:%d' % self.target_object.id.val)
+
+        parameters.page(0, 1)
+        self.target_object = unwrap(query_service.findByQuery(
+            'select i from Image as i where i.id = :id',
+            parameters, {'omero.group': '-1'}))
+        self.target_name = self.target_object.name.val
+
+        data = list()
+        while True:
+            parameters.page(len(data), 1000)
+            rv = query_service.findAllByQuery((
+                'select distinct r from Image as i '
+                'join i.rois as r '
+                'where i.id = :id order by r.id desc'),
+                parameters, {'omero.group': '-1'})
+            if len(rv) == 0:
+                break
+            else:
+                data.extend(rv)
+        if not data:
+            raise MetadataError('Could not find target object!')
+
+        rois_by_id = dict()
+        for roi in data:
+            rid = roi.id.val
+            rois_by_id[rid] = roi
+        self.rois_by_id = rois_by_id
+        log.debug('Completed parsing image: %s' % self.target_name)
+
+
 class ParsingUtilFactory(object):
 
     def get_filter_for_target(self, column_index, target_name):
@@ -905,6 +972,8 @@ class ParsingContext(object):
             return DatasetAnnotationLinkI()
         if ProjectI is self.target_class:
             return ProjectAnnotationLinkI()
+        if ImageI is self.target_class:
+            return ImageAnnotationLinkI()
         raise MetadataError(
             'Unsupported target object class: %s' % self.target_class)
 
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -335,6 +335,7 @@ class ValueResolver(object):
         self.target_class = self.target_object.__class__
         self.target_type = self.target_object.ice_staticId().split('::')[-1]
         self.target_id = self.target_object.id.val
+        self.ambiguous_naming = False
         q = "select x.details.group.id from %s x where x.id = %d " % (
             self.target_type, self.target_id
         )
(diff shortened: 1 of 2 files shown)


sbesson:
Closing/reopening to use the new GitHub action

src/omero_metadata/populate.py
```diff
@@ -146,6 +147,10 @@ class HeaderResolver(object):
 
     DEFAULT_COLUMN_SIZE = 1
 
+    image_keys = {
+        'roi': RoiColumn
+    }
+
```
sbesson, line 153:
```suggestion

```

src/omero_metadata/populate.py
```diff
@@ -838,6 +853,58 @@ def _load(self):
         log.debug('Completed parsing project: %s' % self.target_object.id.val)
 
 
+class ImageWrapper(ValueWrapper):
+
+    def __init__(self, value_resolver):
+        super(ImageWrapper, self).__init__(value_resolver)
+        self.rois_by_id = dict()
+        self._load()
+
```
sbesson, line 862:
```suggestion

```

src/omero_metadata/populate.py
```diff
@@ -838,6 +852,57 @@ def _load(self):
         log.debug('Completed parsing project: %s' % self.target_object.id.val)
 
 
+class ImageWrapper(ValueWrapper):
+
+    def __init__(self, value_resolver):
+        super(ImageWrapper, self).__init__(value_resolver)
+        self.rois_by_id = dict()
+        self._load()
+    def resolve_roi(self, column, row, value):
```
sbesson, line 861:
```suggestion

    def resolve_roi(self, column, row, value):
```

sbesson:
Thanks for the contribution @erindiel. I successfully tested with a very minimal example:

Screenshot 2020-11-09 at 11 36 48
Screenshot 2020-11-09 at 11 37 09

Confirmed that the population script works as expected and creates an OMERO.table that is accessible via the client. If additional rows with non-existent ROIs are passed to the image, these rows are stripped from the table like the rest of the metadata population worfklow.

The API addition makes sense although it could benefit from a minimal integration test to prevent future regression.

A few general thoughts and questions:

- in terms of UI cross-referencing, similarly to how Well and Image columns are linking to the relevant `?show=..` URL, I would almost like to see the ROIs above linking to the  `/iviewer/?roi=413362` URL introduced in https://github.com/ome/omero-iviewer/pull/311. I understand the latter endpoints are specific to `iviewer` but would there a way to forward a webclient call @will-moore  ?
- is it worth looking into a minimal integration test for the new functionality to prevent future regression?
- the worklow described above requires an a priori knowledge of the ROI IDs. My assumption is that it is driven by the use case where images and ROIs have been created on a server and derived analytics are run against it. A separate use case, that we are getting more on the publication/IDR side, is the one where ROIs with associated metadata have been pre-computed and would both need to be populated. In that case, a naive assumption is that the workflow above should be extendable to support and hande a `roi name` column similarly to `image` vs `image name`. Have you faced this use case and/or some experience with this use case?

manics:
> I would almost like to see the ROIs above linking to the /iviewer/?roi=413362 URL 

If Iviewer is the default viewer it's loaded using the default `/weclient/img_detail/<iid>` endpoint:
https://idr.openmicroscopy.org/webclient/img_detail/10560145/
So maybe you could pass additional query parameters `?roi=413362` and have them forwarded to whatever viewer is configured? Effectively this defines a new API that implementers of `/weclient/img_detail/` should support (though they are of course free to ignore parameters)

will-moore:
If iviewer is opened via `/webclient/img_detail/18188/` or `/iviewer/?images=18188` then in either case we have the Image ID, so that adding `?shape=123` is ignored.
However, there is the makings of an API in that iviewer and the old viewer support e.g. rendering settings `?c=-1|0:2417.9873046875$0000FF` etc. Also iviewer and pathviewer (I think) support `?dataset=3183`.

erindiel:
With 7c162daed7ed3088d3d815f0193a2d4b29c31b72, you should be able to use either an `roi` type column with ROI ID or `s` type column with `Roi Name`. The other will be appended automatically. If you use both, nothing will be appended. 25391af929012287967f753a8f79732d5b90b9fa adds some documentation of this.

96e9308394db3cc83c3292e188347b7a304cc901 errors if ambiguous names are used as the identifying column (ambiguous names are OK if IDs are there too).

Open to feedback on whether this covers your use cases @sbesson!

src/omero_metadata/populate.py
```diff
@@ -1252,6 +1379,39 @@ def post_process(self):
...
+                    log.debug(roi_name_column)
+                    rname = roi_name_column.values[i]
+                    rid = self.value_resolver.get_roi_id_by_name(rname)
+                except KeyError:
+                    log.warn(
+                        "%d not found in roi names" % rname)
+                assert i == len(roi_column.values)
+                roi_column.values.append(rid)
+            else:
+                log.info('No ROI information resolution needed, skipping.')
```
sbesson, line 1413:
```suggestion
                log.debug('No ROI information resolution needed, skipping.')
```

As INFO is the default debugging level, having this line printed every row if you have a CSV with ROI ID and names populated is a bit too verbose.

src/omero_metadata/populate.py
```diff
@@ -1252,6 +1379,39 @@ def post_process(self):
...
+                        "%d not found in roi ids" % rid)
+                assert i == len(roi_name_column.values)
+                roi_name_column.values.append(rname)
+                roi_name_column.size = max(
+                    roi_name_column.size, len(rname))
+            elif roi_name_column is not None and (
+                    ImageI is target_class and
+                    resolve_roi_ids and not resolve_roi_names):
+                if self.value_resolver.ambiguous_naming:
+                    raise MetadataError('Cannot resolve ROI names.')
```
sbesson, line 1401: Big :+1: for this logic introduced in 96e9308394db3cc83c3292e188347b7a304cc901. On the IDR front, this is one of the checks we are performing externally on more traditional screen/project table populations.

I assume it would make sense to implement it in other scenarios when e.g. Dataset/Image or Plate/Well cannot be resolved too in order to error the metadata population script with a meaningful error message. /cc @pwalczysko @dominikl

sbesson (approved):
Thanks for the additional commits. Tested a few scenarios using on of the IDR use cases where point ROIs are created from tabular data for a series of images.

I adjusted the ROI creation script to return a 1 CSV per image containing the original tabular metadata incl. the position information as well as the ID and/or the name of the created ROI e.g.

```csv
Roi,Sequence,Tagged Protein,Sequence is unique,Location_X,Location_Y
1991307,CCATCTT,EED_Intron1_1,YES,2864.929563,82.84913737
1991308,GACCAAG,GK5_Intron12_1,YES,2417.862895,141.6575514
1991309,ATAACCT,OXCT1_Intron15_1,YES,5099.678363,189.2736923
1991310,CCATCTT,EED_Intron1_1,YES,2825.886615,154.5139825
```

Note in this use case, as you can see above, the `Tagged Protein` used for the ROI name is non-unique across ROIs.

- `Roi` column only:

Screenshot 2020-11-12 at 14 02 52

- `Roi Name`: the population script errored as expected

```
bash-4.2$ venv/bin/omero metadata populate Image:10508901 --file 1050890 
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
WARNING:omero_metadata.populate:Conflicting ROI names.
INFO:omero_metadata.populate:Created new table OriginalFile:31697737
  File "venv/bin/omero", line 11, in <module>
    sys.exit(main())
  File "/tmp/venv/lib64/python3.6/site-packages/omero/main.py", line 125, in main
    rv = omero.cli.argv()
  File "/tmp/venv/lib64/python3.6/site-packages/omero/cli.py", line 1784, in argv
    cli.invoke(args[1:])
  File "/tmp/venv/lib64/python3.6/site-packages/omero/cli.py", line 1222, in invoke
  File "/tmp/venv/lib64/python3.6/site-packages/omero/cli.py", line 1299, in onecmd
  File "/tmp/venv/lib64/python3.6/site-packages/omero/cli.py", line 1381, in execute
    args.func(args)
  File "/tmp/venv/lib64/python3.6/site-packages/omero_metadata/cli.py", line 505, in populate
    ctx.parse()
  File "/tmp/venv/lib64/python3.6/site-packages/omero_metadata/populate.py", line 1109, in parse
    return self.parse_from_handle_stream(f2)
  File "/tmp/venv/lib64/python3.6/site-packages/omero_metadata/populate.py", line 1076, in parse_from_handle_stream
    self.filter_function, table, 1000)
  File "/tmp/venv/lib64/python3.6/site-packages/omero_metadata/populate.py", line 1205, in populate_from_reader
    self.post_process()
  File "/tmp/venv/lib64/python3.6/site-packages/omero_metadata/populate.py", line 1401, in post_process
    raise MetadataError('Cannot resolve ROI names.')
omero_metadata.populate.MetadataError: Cannot resolve ROI names.
```

- `Roi` and `Roi Name` columns

Screenshot 2020-11-12 at 14 12 35

Overally, everything behaved as expected. The only question in the second scenario is whether we could avoid the creation of a table at all if there is some duplicity in the primary keys of the CSV. 

Note this is a conversation larger than this feature as it would be beneficial to the other metadata population scenarios.

From my side, happy to see this merged and tagged as a minor release.

PR #54, opened 2021-02-15
Close callbacks

joshmoore:
If an error is thrown during submit() and closehandle is true,
then the callback is automatically closed. If not, then the
caller is required to call `cb.close(True)` which will remove
the callback servant. If this is *not* done, then the following
type of error will be thrown on map annotation deletion:

```
    serverExceptionClass = "ome.conditions.OverUsageException"
    message = "servantsPerSession reached for 36cb8357-997a-46ff-b421-5b08d97cd3fc: 10000"
````


sbesson (approved):
Straightforward resource closing change. Release immediately as `0.7.1`?

joshmoore:
Works me. I'll tackle tomorrow AM if you don't beat me to it.

PR #58, opened 2021-05-31
Roi name allowed none

will-moore:
In the case when we `populate metadata` on a target Image with a csv as in the README (below), the ROI IDs are already in the csv and we don't need the ROI.name in OMERO to identify which row applies to which ROI.
However, without this PR, any ROIs without the `name` set in OMERO would cause `populate metadata` to fail.

```
    # header roi,l,d,l
    Roi,Shape,probability,area
    501,1,0.8,250
    502,1,0.9,500
    503,1,0.2,25
```

To test:
 - Draw some Shapes on an Image in iviewer and Save
 - Note the ROI IDs from iviewer and update the ROIs in a csv like the one above.
 - Run `populate metadata` script on the Image in webclient with the csv (this requires https://github.com/ome/omero-scripts/pull/184) or use the CLI.
 - Should get an OMERO.table created on the Image with a `Roi Name` column where all the values will be empty
 - Bonus: If you also include a `Shape` column of type `l` with the correct Shape values, it should be possible to use this OMERO.table to create a heatmap on those Shapes in OMERO.figure as described in https://github.com/ome/omero-guide-figure/pull/11 

cc @pwalczysko

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,7 +147,7 @@ If the target is an Image, a csv with ROI-level and object-level data can be use
 The ROI identifying column can be an ``roi`` type column containing ROI ID, and ``Roi Name``
 column will be appended automatically (see example below). Alternatively, the input column can be
 ``Roi Name`` (with type ``s``), and an ``roi`` type column will be appended containing ROI IDs.
-In *both* cases, it is required that ROIs on the Image in OMERO have the ``Name`` attribute set.
+In this case, it is required that ROIs on the Image in OMERO have the ``Name`` attribute set.
 
 image.csv::
 
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -896,7 +896,7 @@ class ImageWrapper(ValueWrapper):
         return self.rois_by_name[rname].id.val
 
     def get_roi_name_by_id(self, rid):
-        return self.rois_by_id[rid].name.val
+        return unwrap(self.rois_by_id[rid].name)
 
     def resolve_roi(self, column, row, value):
         try:
@@ -940,10 +940,10 @@ class ImageWrapper(ValueWrapper):
         for roi in data:
             rid = roi.id.val
             rois_by_id[rid] = roi
-            if roi.name.val in rois_by_name.keys():
+            if unwrap(roi.name) in rois_by_name.keys():
                 log.warn('Conflicting ROI names.')
                 self.ambiguous_naming = True
-            rois_by_name[roi.name.val] = roi
+            rois_by_name[unwrap(roi.name)] = roi
         self.rois_by_id = rois_by_id
         self.rois_by_name = rois_by_name
         log.debug('Completed parsing image: %s' % self.target_name)
@@ -1397,6 +1397,8 @@ class ParsingContext(object):
                     log.debug(roi_column)
                     rid = roi_column.values[i]
                     rname = self.value_resolver.get_roi_name_by_id(rid)
+                    if rname is None:
+                        rname = ""
                 except KeyError:
                     log.warn(
                         "%d not found in roi ids" % rid)


will-moore:
Seems ALL tables tests failed with:
```
    def newTable(self, repoId, path, _ctx=None):
>       return _M_omero.grid.SharedResources._op_newTable.invoke(self, ((repoId, path), _ctx))
E       omero.InternalException: exception ::omero::InternalException
E       {
E           serverStackTrace = 
E           serverExceptionClass = 
E           message = null table as argument
E       }
```
as if `pytables` install failed or something?

joshmoore:
It could have either fallen over today (happens very rarely) or a code bug in an open PR prevented it from starting.

will-moore:
@pwalczysko The error I see there is:

```
Column type '    roi' unknown.
Choose from following: plate,well,image,dataset,roi,d,l,s,b
```

So it looks like you've got some extra whitespace in the header. 

I just tried with image https://merge-ci.openmicroscopy.org/web/webclient/?show=image-83221 and the csv below:

```
# header roi,l,d,l
Roi,Shape,probability,area
273462,279241,0.8,250
```

using the webclient to pick the csv from the script dialog. You can see that table with 1 row was created: https://merge-ci.openmicroscopy.org/web/webclient/omero_table/649358/

NB: this also tests that https://github.com/ome/omero-scripts/pull/184 is working.

pwalczysko:
thank you @will-moore , there was a tab preceding the first hash and the second line starting with "Roi". I removed the tab from the csv and it started to work. I can see that the table has nothing in the ROI name column as indicated. https://merge-ci.openmicroscopy.org/web/webclient/omero_table/649360/

This looks good to me.

pwalczysko:
The "bonus" workflow fails for me though. 
The script in Console produces a link https://merge-ci.openmicroscopy.org/webgateway/table/Image/144515/query/?query=Shape-480591 which gives 404. But the ROI indeed has the correct shape as verified on https://merge-ci.openmicroscopy.org/web/iviewer/?roi=425062 .  It seems to me that should work, but it does not...

will-moore:
Ah, I see - Because merge-ci is deployed with the `/web/` URL prefix, that needs be added to the URL in the script.
to give:
https://merge-ci.openmicroscopy.org/web/webgateway/table/Image/144515/query/?query=Shape-480591

With that change, and loading `probability` column, that gave me heatmap on shapes at:
https://merge-ci.openmicroscopy.org/web/figure/file/76582

I guess I could try to expand the code snippet to handle the presence of the URL prefix automatically, but I don't want to make it too verbose when it's not needed on workshop server etc.
<details><summary>Expand for Full script...</summary>

```
async function shapeData(panel) {
    const shapeIds = panel.get("shapes").map(s => s.id).filter(id => id > 0);
    console.log('shapeIds', shapeIds);
    let vals_by_shape = {};
    for (let i = 0; i < shapeIds.length; i++) {
        // Load one at a time - more reliable
        let url = `/web/webgateway/table/Image/${panel.get('imageId')}/query/?query=Shape-${shapeIds[i]}`;
        let r = await fetch(url).then(rsp => rsp.json());
        let colIndex = r.data?.columns?.indexOf("probability");
        let shapeIndex = r.data?.columns?.indexOf("Shape");
        if (colIndex && shapeIndex && r.data?.rows.length > 0) {
            console.log("Value", r.data.rows[0][colIndex]);
            vals_by_shape[r.data.rows[0][shapeIndex]] = r.data.rows[0][colIndex];
        }
    };
    // Once all loaded, we can calculate range and assign colours to shapes
    const values = Object.values(vals_by_shape);
    let minVal = Math.min(...values);
    let valRange = Math.max(...values) - minVal;
    console.log('min, range', minVal, valRange);
    const new_shapes = panel.get("shapes").map(shape => {
        // hide any shapes we don't have data for
        if (!vals_by_shape[shape.id]) return { ...shape, strokeWidth: 0.01 };
        let value = (vals_by_shape[shape.id] - minVal) / valRange;
        let red = parseInt(value * 255).toString(16);
        let green = (255 - parseInt(value * 255)).toString(16);
        // if e.g. "f", -> "0f"
        red = red.length == 1 ? `0` + red : red;
        green = green.length == 1 ? `0` + green : green;
        return { ...shape, strokeColor: `#${red}${green}00`, strokeWidth: 5 }
    });
    panel.set('shapes', new_shapes);
}
figureModel.getSelected().forEach(shapeData);
```

</details>

will-moore:
OK - this is getting me confused because the "Bonus" was just to complete the figure workflow, but is not related to this PR, which is otherwise good to merge.
So we should move the OMERO.figure scripting discussion to https://github.com/ome/omero-guide-figure/pull/11

pwalczysko:
the funclionality indeed works as described, and the "bonus" test is actually working now too. Ready to merge fmpov.

sbesson:
Overall the proposed change makes sense to me. In the case where ROIs are specified by ID, the requirement for a non-null `name` attribute seems unnecessary as the resolution should be unambiguous. @erindiel can you confirm relaxing this requirement in this scenario will not cause issues in your workflows?

erindiel:
Thanks @will-moore and @sbesson . Agreed that the current requirement for a non-null ROI Name should not be the case when ROI ID is provided. Thanks for the change, and no issues for our workflows.

Issue #59, opened 2021-07-16
Regression caused by NaN support


sbesson:
With omero-metadata 0.8.1, trying to run `omero metadata populate --context bulkmap` fails with

```
[sbesson@pilot-idr0113-omeroreadwrite ~]$ /opt/omero/server/venv3/bin/omero metadata populate --context bulkmap Screen:2801  --cfg /uod/idr/metadata/idr0097-reicher-proteintag/screenA/idr0097-screenA-bulkmap-config.yml 
Using session for public@localhost:4064. Idle timeout: 10 min. Current group: Public
Traceback (most recent call last):
  File "/opt/omero/server/venv3/bin/omero", line 11, in <module>
    sys.exit(main())
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/main.py", line 125, in main
    rv = omero.cli.argv()
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1784, in argv
    cli.invoke(args[1:])
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1222, in invoke
    stop = self.onecmd(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1299, in onecmd
    self.execute(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1381, in execute
    args.func(args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_metadata/cli.py", line 536, in populate
    allow_nan=args.allow_nan)
TypeError: __init__() got an unexpected keyword argument 'allow_nan'
```

This regression comes from the NaN support introduced in https://github.com/ome/omero-metadata/pull/53 where a new `allow_nan` key/value pair is defined in the constructor of `ParsingContext` and set via the command-line `--allow-nan` flag.

Since the other Context objects have not been updated with the same change, passing any non-default  context e.g. `--context bulkmap` or `--context deletemap` will currently fail at initialization time with the `TypeError` above.

An immediate workaround would be to add an `allow_nan=False` to the constructor of both context objects. Since this will likely be a no-op for the purpose of the other contexts, an alternative would be to use `**kwargs` in the constructor to handle an arbitrary number of arguments.

/cc @jburel

sbesson:
Closed by #60

PR #60, opened 2021-07-19
add allow_nan flag to context

jburel:
The problem appears when I ran 
```
omero metadata populate --context bulkmap --batch 100 --cfg idr0113-experimentA-bulkmap-config.yml Project:1855
```

the change was tested in the context above
```
omero metadata populate --context bulkmap --batch 100 --cfg idr0113-experimentA-bulkmap-config.yml Project:1855
_save_annotation_and_links with batch size 100
INFO:omero_metadata.populate:Created/linked 110 MapAnnotations (total 110)
INFO:omero_metadata.populate:Created/linked 100 MapAnnotations (total 210)
INFO:omero_metadata.populate:Created/linked 10 MapAnnotations (total 220)
```

cc @sbesson

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -1545,7 +1545,8 @@ class BulkToMapAnnotationContext(_QueryContext):
 
     def __init__(self, client, target_object, file=None, fileid=None,
                  cfg=None, cfgid=None, attach=False, options=None,
-                 batch_size=1000, loops=10, ms=10, dry_run=False):
+                 batch_size=1000, loops=10, ms=10, dry_run=False,
+                 allow_nan=False):
         """
         :param client: OMERO client object
         :param target_object: The object to be annotated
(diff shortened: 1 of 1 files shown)


dominikl:
With latest version 0.8.1 I get`TypeError: __init__() got an unexpected keyword argument 'allow_nan'`, with the PR it works fine. Looks good 👍

src/omero_metadata/populate.py
```diff
@@ -1545,7 +1545,8 @@ class BulkToMapAnnotationContext(_QueryContext):
 
     def __init__(self, client, target_object, file=None, fileid=None,
                  cfg=None, cfgid=None, attach=False, options=None,
-                 batch_size=1000, loops=10, ms=10, dry_run=False):
+                 batch_size=1000, loops=10, ms=10, dry_run=False,
+                 allow_nan=False):
```
sbesson, line 1549: This will definitely do the job to fix the regression. Thanks for releasing.
As mentioned in #59, we might need to think of a strategy that does not involve adding an no-op keyword for every single `Context` class every time a feature is implemented.

PR #61, opened 2021-08-18
Add table_name option to ParsingContext()

will-moore:
See https://github.com/IDR/idr0079-hartmann-lateralline/pull/5#issuecomment-901249131

Usage:
```
    ctx = ParsingContext(
        client, image._obj, file=csv_name, allow_nan=True,
        table_name=table_name
    )
    ctx.parse()
```
This change used in https://github.com/IDR/idr0079-hartmann-lateralline/pull/5/commits/57e507fe24f34170d5b60a91e709ff6218bfbabd
to create bulk annotations with custom names:

Screenshot 2021-08-18 at 22 28 56

NB: `ParsingContext()` has several parameters that are not used: `fileid`, `cfg`, `cfgid`, `attach`, `options`, `batch_size`, `loops`, `ms`. Seems like these should be removed, although that will break code that still includes them.

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -1013,6 +1013,8 @@ class ParsingContext(object):
                                                        target_object,
                                                        self.value_resolver)
         self.dry_run = dry_run
+        if not table_name:
+            table_name = DEFAULT_TABLE_NAME     # just in case
         self.table_name = table_name
 
     def create_annotation_link(self):
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -988,7 +988,7 @@ class ParsingContext(object):
     def __init__(self, client, target_object, file=None, fileid=None,
                  cfg=None, cfgid=None, attach=False, column_types=None,
                  options=None, batch_size=1000, loops=10, ms=500,
-                 dry_run=False, allow_nan=False):
+                 dry_run=False, allow_nan=False, table_name=None):
         '''
         This lines should be handled outside of the constructor:
 
@@ -1012,6 +1012,7 @@ class ParsingContext(object):
                                                        target_object,
                                                        self.value_resolver)
         self.dry_run = dry_run
+        self.table_name = table_name
 
     def create_annotation_link(self):
         self.target_class = self.target_object.__class__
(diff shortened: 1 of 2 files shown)


src/omero_metadata/populate.py
```diff
@@ -988,7 +988,7 @@ class ParsingContext(object):
     def __init__(self, client, target_object, file=None, fileid=None,
                  cfg=None, cfgid=None, attach=False, column_types=None,
                  options=None, batch_size=1000, loops=10, ms=500,
-                 dry_run=False, allow_nan=False):
+                 dry_run=False, allow_nan=False, table_name=None):
```
sbesson, line 991: Any reason not to use `, table_name=DEFAULT_TABLE_NAME` here (and remove the conditional below)?

sbesson (commented):
One minor question about the default value. Otherwise this change looks fairly straightforward.

As mentioned in the related PR, I did not find any  downstream code that relies on the table name and would be broken by this PR. Instead all downstream code filters by namespace.


Thanks for raising the unused parameters. As it stands, the issue is that `omero metadata populate` will pass arguments independently of the selected context so there is a need for some redundant declaration at the moment.
 In https://github.com/ome/omero-metadata/issues/59#issue-946613849, I suggested that using `**kwargs` might be an option to be lenient about extra parameters and declare only the relevant parameters across `Context` implementations.

src/omero_metadata/populate.py
will-moore: Done in 052b753b3f

sbesson:
Thanks. Only two questions are:
- can we add an integration test to capture the new keyword?
- after that, should we release as `0.9.0` as this is a new backwards compatible API?

will-moore:
Added tests

test/integration/metadata/test_populate.py
```diff
@@ -1018,27 +1020,33 @@ def test_get_config_remote(self):
...
             attach to Plate. Then query to check table has expected content.
         """
 
         target = fixture.get_target()
+        kwargs = {}
+
         allow_nan = None
         if hasattr(fixture, 'allow_nan'):
             allow_nan = fixture.allow_nan
+        kwargs['allow_nan'] = allow_nan
```
sbesson, line 1035: Nice usage of `kwargs`. Could this call be moved under the `if hasattr(fixture, 'allow_nan'):` block above so that we also test the scenario whenn `allow_nan` is not passed as a keyword?

sbesson (approved):

sbesson:
Now released as https://libraries.io/pypi/omero-metadata/0.9.0

PR #62, opened 2021-11-26
Support ROI table on Dataset, for table with ROI ID column

will-moore:
This supports work on idr0123 where we are looking to create a Table with ROIs linked to a Dataset.

The table has an `roi` column so we know the ROI IDs already and we don't load any ROIs or add an `Roi Name` column.

NB: Quite often in these workflows, I am creating a csv table that has all the data I need (don't need to do any resolving between object names), and it feels quite hard to simply turn this table into an OMERO table with `populate.py` because it's designed to do object name resolution. So maybe I'm not using the right tool for the job.
Although the above changes are quite small, they took quite a while to figure out!

E.g. I really just want a `create_table(csv)` command that simply creates an OMERO.table from the csv as-is, possibly using a `#header` of column types, but not doing any other look-ups etc. I think this has also been requested by users

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -253,6 +253,7 @@ class HeaderResolver(object):
         return self._create_columns("image")
 
     def _create_columns(self, klass):
+        target_class = self.target_object.__class__
         if self.types is not None and len(self.types) != len(self.headers):
             message = "Number of columns and column types not equal."
             raise MetadataError(message)
@@ -308,7 +309,7 @@ class HeaderResolver(object):
                               self.DEFAULT_COLUMN_SIZE, list()))
                 # Ensure ImageColumn is named "Image"
                 column.name = "Image"
-            if column.__class__ is RoiColumn:
+            if column.__class__ is RoiColumn and target_class != DatasetI:
                 append.append(StringColumn(ROI_NAME_COLUMN, '',
                               self.DEFAULT_COLUMN_SIZE, list()))
                 # Ensure RoiColumn is named 'Roi'
@@ -758,6 +759,10 @@ class DatasetWrapper(PDIWrapper):
         self.images_by_id = dict()
         self.images_by_name = dict()
         self._load()
+
+    def resolve_roi(self, column, row, value):
+        # Support Dataset table with known ROI IDs
+        return int(value)
 
     def get_image_id_by_name(self, iname, dname=None):
         return self.images_by_name[iname].id.val


sbesson:
Still brainstorming on this, have you considered including an `Image` or `Image Name` column that would need to be resolved as well primarily for consistency? A downside of the structure proposed here is that unlike in the Image-level table. the Image/ROI relationship is lost. I assume that if a downstream application wanted to query a table e.g. find all spots associated with a range of a chromosome across a datasets, this would be an useful information.

will-moore:
Yes, good point. But I would be reluctant to add the logic for `Image` lookup to `populate.py`. You'd need a big look-up to load all the Images for all the ROIs in the Dataset, and the permutations of the code are hard to understand. But it would be trivial for me to add them to my csv when I'm creating it. So this is another case where I'd just want a simple csv -> Table.

sbesson:
For prototyping, I definitely see the value of being able to do CSV -> OMERO.tables and maybe there is a flag that just does so.
But in production, I consider the lookup is one of the features of the omero-metadata plugin in order to validate the containment before linking CSV columns to Image, Dataset, Well... typed columns. 

Is the Image/ROI lookup significantly more expesnse than e.g. a Project/Dataset/Image lookup or a Screen/Plate/Well lookup? Or is the issue that the code path is not in the plugin yet?

will-moore:
@sbesson This now supports look-up of ROIs and Shapes for a DatasetWrapper.
It uses the same behaviour as for ImageWrapper: returns a `Skip()` if the ROI or Shape isn't found.
This is different than handling of Images within a Dataset, where we return `-1` if the Image ID isn't found.
Also added a test for this behaviour.

README.rst
```diff
@@ -142,36 +170,45 @@ Well  Plate  Drug   Concentration  Cell_Count  Percent_Mitotic  Well Name   Plat
 
 If the target is a Plate instead of a Screen, the ``Plate`` column is not needed.
 
-If the target is an Image, a csv with ROI-level and object-level data can be used to create an
-``OMERO.table`` (bulk annotation) as a ``File Annotation`` on an Image.
+**ROIs**
+
+If the target is an Image or a Dataset, a csv with ROI-level and Shape-level data can be used to create an
```
jburel, line 175: unify to CSV

src/omero_metadata/populate.py
```diff
@@ -757,8 +760,37 @@ def __init__(self, value_resolver):
         super(DatasetWrapper, self).__init__(value_resolver)
         self.images_by_id = dict()
         self.images_by_name = dict()
+        self.rois_by_id = None
+        self.shapes_by_id = None
         self._load()
 
+    def resolve_roi(self, column, row, value):
+        # Support Dataset table with known ROI IDs
+        print("DatasetWrapper resolve_rois", value)
```
jburel, line 769: remove or add to ``log``

src/omero_metadata/populate.py
```diff
@@ -806,6 +838,43 @@ def _load(self):
...
+                'join i.datasetLinks as dil '
+                'join dil.parent as d '
+                'where d.id = :id order by s.id desc'),
+                parameters, {'omero.group': '-1'}))
+            if len(rv) == 0:
+                break
+            else:
+                data.extend(rv)
+        if not data:
+            print("No ROIs on images in target Dataset")
```
jburel, line 865: add to log too?

jburel:
Test 1:
Link to Project using similar example that the one in README. The dataset contains 4 images. Only 2 listed in the CVS and one that is not in the dataset is listed.
The output of the command is a unexpected i.e.
```
INFO:omero_metadata.populate:Adding dataset:3710 image:49209
INFO:omero_metadata.populate:Adding dataset:3710 image:49208
INFO:omero_metadata.populate:Adding dataset:3710 image:49207
INFO:omero_metadata.populate:Adding dataset:3710 image:49206
```
From the output, I thought before checking the table that the entries for the 4 images have been added to the table.

The choice of ``-1`` for the image option means that the web display still offers an "hyperlink" option
Screenshot 2022-01-12 at 13 51 46

jburel:
Test: omero metadata populate Dataset:3710 --file dataset.csv 
No output
Expecting something like ``INFO:omero_metadata.populate:Adding image:49209`` to match the previous command

According to the README ``If the target is a Dataset instead of a Project, the ``Dataset Name`` column is not needed.``
* If the dataset name is specified and is not valid, the image is not considered. There is in that case no entry in the table, even if the image is in the dataset specified via the command line (i.e. 3710). This is inconsistent with the Project command since a ``-1`` was inserted.
* If there is no dataset name, the image is added to the table

jburel:
Test ```if you already know the Image IDs, these can be specified
in an ``image`` column, and an ``Image Name`` column will be added.
This is only supported for a Dataset.```

* If all IDs are valid, the image names are set.
* If an ID is not the ID of an image in the Dataset, the ID in the Image column is set to ``-1`` and no image name is set

will-moore:
@jburel Apologies for the confusion, but the scenarios you're testing there aren't ones that have been added in this PR.
This PR adds support for tables with an `roi` column (and `shape` column) with roi/shape IDs, where the target is a Dataset e.g:

```
    # header roi,l,l,d,l
    Roi,shape,object,probability,area
    501,1066,1,0.8,250
    502,1067,2,0.9,500
    503,1068,3,0.2,25
```

jburel:
Test ``omero metadata populate Image:49206 --file rois.csv``
* The ROis have no name. Message ``WARNING:omero_metadata.populate:Conflicting ROI names`` is misleading. 
* One roi ID in the CSV is incorrect ``WARNING:omero_metadata.populate:Image is missing ROI: 1286504`` shows up twice in the output. The noise is not really necessary. This results with no entry in the table. This is again inconsistent with the previous methods.

jburel:
``'ImageWrapper' object has no attribute 'resolve_shape'``

will-moore:
@jburel I moved the README changes that were unrelated to this PR to #66.
I'll add support for `shape` column to `ImageWrapper` since it is confusing to only have that for Dataset.

src/omero_metadata/populate.py
will-moore: This can revert to the `raise MetadataError()` that is used elsewhere. I didn't use that initially as `_load_rois()` was called even if there were no `shape` or `roi` column in the table, but now it is only called when needed so it makes sense to throw.

will-moore:
@jburel Added support for `shape` column for target `Image:ID`. Also added that to tests.

pwalczysko:
```
 If a column named shape of type l is included, the Shape IDs will be validated and if an image ID column is included, an Image Name column will be added as above.
```
A sentence from the Readme - I am afraid I do not understand it. Does it really want to say that I have to have ``shape`` column of type long and only if this condition is fullfilled, then, if additionally an image ID column is included, an Image Name column will be added ? This seems to me the meaning of the very long sentence, but hard to believe this is accurrate. Also, what does ``above`` refer to, I cannot see any ``Image Name`` column anywhere above ?

will-moore:
That sentence could be split up which might help:

No you don't have to have a shape column.
`If a column named shape of type l is included (as in the example), the Shape IDs will be validated`
which currently means that if the shape IDs are invalid then the row won't get included. I haven't explained handling of invalid values here since we don't do it anywhere else yet and it adds to the complexity of the doc. But it might be worth doing it.

`If an image ID column is included, an Image Name column will be added as above.
The Image IDs -> Image Name workflow that was above got moved to a different PR, since it was describing functionality that wasn't added in this PR: https://github.com/ome/omero-metadata/pull/66/files

will-moore:
Re https://github.com/ome/omero-metadata/pull/62#issuecomment-1011216448
cc @erindiel - Do you have strong feelings on the handling of invalid ROI IDs? In this PR I've adopted your logic of using a `Skip` if the ROI ID is invalid in `resolve_roi()` so the whole CSV row gets ignored, but this is different from most other handling where we use `-1` if e.g. the `Roi Name` or `Image Name` is not found.
So I'm thinking of using `-1` and keeping the row for ROI IDs and Shape IDs, but don't know if you're relying on `Skip()`?

will-moore:
@erindiel Let me know if keeping a row with `-1` for invalid ROI IDs would cause problems, since I think that's what I'd like to go for (to be consistent with handling Image IDs etc). Thanks.

will-moore:
@jburel This now handles invalid ROI and Shape IDs by setting them to -1 instead of skipping the row.
@pwalczysko Tweaked the README to try and make it clearer.

will-moore:
Used this in it's current form to re-create OMERO.tables on Datasets in `idr0123` https://idr-redmine.openmicroscopy.org/issues/270#note-13

pwalczysko:
>  @pwalczysko Tweaked the README to try and make it clearer.

Thank you @will-moore . I still have following questions regarding the ROIs paragraph.

1. 
Screenshot 2022-01-19 at 18 18 30
I would still think that it would be better to say what examples are meant. Are those the ROI and Shape examples from the same paragraph ? The "above" is really not reassuring.

2. 
Screenshot 2022-01-19 at 18 27 41
Futher in the same sentence, the formatting of the words "Image ID" is confusingly different from the ROI ID formatting just two sentences above. Maybe rather say ``image`` column containing IDs, to avoid any doubt ?


3. Where are the ROI names taken from ? If I have some ROIs in OMERO and somehow retrieve their IDs, create a CSV and then try to populate it with results and import the CSV to OMERO using this metadata plugin, I do not suppose that the ROIs are by default named ? What will the plugin do then ? Error ? How to prevent ?

4. The fact that ``shape`` type is not supported means that I cannot target Shapes directly in my CSV, they always have to be "under" a ROI which is specified in the same line of my CSV ?

erindiel:
> @erindiel Let me know if keeping a row with `-1` for invalid ROI IDs would cause problems, since I think that's what I'd like to go for (to be consistent with handling Image IDs etc). Thanks.

@will-moore we are not relying on skipping the row if the ROI ID is invalid, so replacing with `-1` to be consistent with the handling of other object IDs works for us. In our workflow, there is typically a single ROI ID across the entire table, and a table with an invalid ROI ID will be "broken" regardless. 👍🏻

will-moore:
Great, thanks

will-moore:
@pwalczysko Tried to address some of your feedback, but some is a bit out of scope for this PR and comes under:
How do we fully document the behaviour of `metadata populate`?
There are so many options: targets, column types etc. that it feels like we need a way to describe how everything is handled.
Probably needs some kind of docs page since it's getting a bit verbose for the README.

pwalczysko:
> @pwalczysko Tried to address some of your feedback, but some is a bit out of scope for this PR and comes under: How do we fully document the behaviour of `metadata populate`? There are so many options: targets, column types etc. that it feels like we need a way to describe how everything is handled. Probably needs some kind of docs page since it's getting a bit verbose for the README.

Thank you.
At the moment, I am simply trying to understand the behaviour myself, so not worried just yet about the user understanding it. Surely we have to go away from the README. I would suggest that we need plenty of examples. Not sure about whether to target guides or docs (and link from guides to the docs ? ) cc @jburel @sbesson 
For that, I still have a question

``If a column named shape of type l is included, the Shape IDs will be validated (and set to -1 if invalid). ``
Does that mean that that column named shape is containing Shape IDs ? If yes, I think the sentence should first clearly state that, only then continue about explaining what will be done with these IDs.

pwalczysko:
@will-moore thank you, now it indeed leaves little room for doubt, I think I understand what to do.

jburel:
I think usage guide will be good. We have section about CLI import already in the guide so the same strategy could be followed for that plugin

will-moore:
Updated the description with testing instructions.

pwalczysko:
> I think usage guide will be good. We have section about CLI import for example already in the guide so the same strategy could be followed for that plugin. Adding it to the README does not scale. I think we agreed to add ref doc in the repository itself a while back.
> 
> ```
> so not worried just yet about the user understanding it
> ```
> 
> That does not sound like a good start.

@jburel we already have https://omero-guides.readthedocs.io/en/latest/upload/docs/metadata-ui.html#create-a-metadata-csv as well as https://omero-guides.readthedocs.io/en/latest/upload/docs/metadata.html
I think we need a new chapter to alleviate the confusion the user is feeling when trying to create a workable CSV.

jburel:
Thanks @pwalczysko I forgot about it.
We have 2 types of doc
* the guide 
* the ref doc. For the ref doc putting the generated and not generated doc from this repo will follow what has been done elsewhere. That could be easily published on rtd

pwalczysko:
Test 2:
created a csv with multiple rois per image and an image column and also multiple shapes per roi. Worked fine, see the csv bleow.

```
# header roi,l,image,l,d,l
roi,shape,image,object,probability,area
437462,526593,140169,1,0.8,250
437462,526594,140169,2,0.11,400
437463,526597,140169,2,0.11,400
437465,526607,140167,3,0.9,500
437469,526621,140168,4,0.2,25
```

pwalczysko:
Test 3
In dataset https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-25469 renamed one image so that I have 2 images with the same name in this dataset. Then, deleted the previous table from [Test2](https://github.com/ome/omero-metadata/pull/62#issuecomment-1042907130) and rerun [Test2](https://github.com/ome/omero-metadata/pull/62#issuecomment-1042907130)  on the same dataset.
This caused an error

<details>

```
omero metadata populate Dataset:25469 --file ~/test-will/imagesroisshapes.csv 
Using session for user-3@merge-ci-devspace.openmicroscopy.org:4064. Idle timeout: 10 min. Current group: read-annotate-1
  File "/Users/pwalczysko/miniconda3/envs/cli/bin/omero", line 11, in <module>
    sys.exit(main())
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/main.py", line 125, in main
    rv = omero.cli.argv()
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1784, in argv
    cli.invoke(args[1:])
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1222, in invoke
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1299, in onecmd
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1381, in execute
    args.func(args)
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/cli.py", line 536, in populate
    allow_nan=args.allow_nan)
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 1094, in __init__
    allow_nan=allow_nan)
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 364, in __init__
    self.wrapper = DatasetWrapper(self)
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 765, in __init__
    self._load()
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 834, in _load
    iname, self.images_by_name[iname].id.val, iid
Exception: Image named Tcf15_KO_d2.tif(id=140168) present. (id=140167)
```

</details>

I would have thought that this Test3 is testing the "identical image names" case mentioned in the last commit ? Does not seem to work.

will-moore:
Ah - Test 1 ROI link is actually an iviewer bug. It's not handling the fact that the ROI is in your non-default group.
I just tried changing the default group for user-3 and logging out. On logging in again the links are working (show ROI in iviewer). I'll create an inviewer issue...

pwalczysko:
As @will-moore aswered the Test1 problem with ROI link above, in summary the remaining problem is https://github.com/ome/omero-metadata/pull/62#issuecomment-1042909191 Test3 (same-named images)

Other than that worked fine.

will-moore:
Test 3 - actually that is showing the fix is working. The bug reported in https://github.com/ome/omero-metadata/issues/68 was that the "duplicate image" error message was failing to render properly. Now it is showing up properly! This doesn't change the fact that duplicate image names aren't supported.

will-moore:
iviewer issue: https://github.com/ome/omero-iviewer/issues/400

pwalczysko (approved):

jburel:
I think we can release as ``0.10.0`` as suggested. @will-moore Do you want to take care of the release?

PR #66, opened 2022-01-12
Add Image IDs workflow to README

will-moore:
Moved REAMDE updates from #62 to a new PR since they were unrelated to the other changes there.

Issue #65, opened 2022-01-10
Support Dataset IDs and Image IDs when target is Project


will-moore:
Currently, this csv works with a target of Dataset, with Image IDs validated and set to -1 if invalid:

```
# header image,d,l,s
Image,ROI_Area,Channel_Index,Channel_Name
1277,0.0469,1,DAPI
1278,0.142,2,GFP-AuroraB
1279,0.093,3,TRITC-edit
1280,0.112233,4,cy5
1,100.1,5,Invalid-Image-ID
```

However, if I try to use a Project as target, with a Dataset column this fails (same result if the `header` column type is `l`):

```
# header image,dataset,d,l,s
Image,Dataset,ROI_Area,Channel_Index,Channel_Name
1277,211,0.0469,1,DAPI
1278,211,0.142,2,GFP-AuroraB
1279,211,0.093,3,TRITC-edit
```

with:

```
  File "/Users/wmoore/Desktop/METADATA/omero-metadata/src/omero_metadata/populate.py", line 1069, in preprocess_from_handle
    self.preprocess_data(reader)
  File "/Users/wmoore/Desktop/METADATA/omero-metadata/src/omero_metadata/populate.py", line 1167, in preprocess_data
    self.post_process()
  File "/Users/wmoore/Desktop/METADATA/omero-metadata/src/omero_metadata/populate.py", line 1343, in post_process
    did = int(columns_by_name["dataset"].values[i])
```

This occurs during the `preprocess_data()`, where we are trying to find the longest Dataset Name, so that the `Dataset Name` column can be created of the correct size (before we populate the table).

Trying the same with HCS data...

With a Screen...
```
# header image,plate,d,l,s
Image,Plate,ROI_Area,Channel_Index,Channel_Name
3559,251,0.0469,1,DAPI
3560,251,0.142,2,GFP-AuroraB
3563,251,0.093,3,TRITC-edit
```

```
$ omero metadata populate Screen:601 --file test-image-plate-ids.csv -vvv
...
  File "/Users/wmoore/Desktop/METADATA/omero-metadata/src/omero_metadata/populate.py", line 1127, in parse
    self.preprocess_from_handle(f1)
  File "/Users/wmoore/Desktop/METADATA/omero-metadata/src/omero_metadata/populate.py", line 1069, in preprocess_from_handle
    self.preprocess_data(reader)
[...]

Issue #51, opened 2020-12-14
Make populate work on Plates


pwalczysko:
When the following command is ueed with a csv as specified below, it ends in error

````
omero metadata populate --context csv --file simple-screen.csv Plate:51
Using session for trainer-1@localhost:4064. Idle timeout: 10 min. Current group: Lab1
Traceback (most recent call last):
  File "/opt/omero/server/venv3/bin/omero", line 118, in <module>
    rv = omero.cli.argv()
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1754, in argv
    cli.invoke(args[1:])
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1187, in invoke
    stop = self.onecmd(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1264, in onecmd
    self.execute(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1346, in execute
    args.func(args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_metadata/cli.py", line 505, in populate
    ctx.parse()
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_metadata/populate.py", line 1107, in parse
    self.preprocess_from_handle(f1)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_metadata/populate.py", line 1049, in preprocess_from_handle
    self.preprocess_data(reader)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_metadata/populate.py", line 1123, in preprocess_data
    column, original_value, row)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_metadata/populate.py", line 448, in resolve
    return self.wrapper.resolve_plate(column, row, value)
AttributeError: 'PlateWrapper' object has no attribute 'resolve_plate'
```

The same command where the target object is a Screen containing the above Plate is successful.

Suggestion: Make the plugin work in both situations, the present behaviour is too strict.

cc @will-moore @sbesson 

The used CSV

```
# header well,plate,s,d,l,d
Well,Plate,Drug,Concentration,Cell_Count,Percent_Mitotic
A1,INMAC384-DAPI-CM-eGFP_59223_1,DMSO,10.1,10,25.4
[...]

Issue #29, opened 2019-07-10
Inconsistent CSV format required for populate metadata on Plate vv Project


will-moore:
Looking at examples added to the README in https://github.com/ome/omero-metadata/pull/28
it seems that the CSV format has different rules for Screen compared with Project:

Screen example has columns of ```Well,Plate``` and the column types are ```well,plate```.
Omitting the ```# header``` has no effect on column types of ```well and plate```.

```
# header well,plate,s,d,l,d
Well,Plate,Drug,Concentration,Cell_Count,Percent_Mitotic
A1,plate01,DMSO,10.1,10,25.4
A2,plate01,DMSO,0.1,1000,2.54
A3,plate01,DMSO,5.5,550,4
B1,plate01,DrugX,12.3,50,44.43
```

This creates an OMERO.table with additional ```Well Name``` and ```Plate Name``` columns and the previous ```Well``` and ```Plate``` columns now have IDs instead of Names.

Well | Plate | Drug | Concentration | Cell Count | Percent Mitotic | Well Name | Plate Name
-- | -- | -- | -- | -- | -- | -- | --
9154 | 3855 | DMSO | 10.1 | 10 | 25.4 | a1 | plate01
9155 | 3855 | DMSO | 0.1 | 1000 | 2.54 | a2 | plate01

If I use column named ```Plate Name``` instead of ```Plate``` in the csv, the script fails:

```
  File "/Users/willadmin/Virtual/omero/lib/python2.7/site-packages/omero_metadata/populate.py", line 1241, in post_process
    plate = columns_by_name['plate'].values[i]   # FIXME
```
and it also fails if I use ```Well Name``` instead of ```Well```.

When the target is Plate, we need columns named ```Dataset Name``` and ```Image Name``` in the CSV and these columns are of type ```String```.

```
# header s,s,d,l,s
Image Name,Dataset Name,Bounding_Box,Channel_Index,Channel_Name
img-01.png,dataset01,0.0469,1,DAPI
img-02.png,dataset01,0.142,2,GFP
img-03.png,dataset01,0.093,3,TRITC
img-04.png,dataset01,0.429,4,Cy5
```
This creates an OMERO.table with an additional ```Image``` ID column:

Image Name | Dataset Name | Bounding_Box | Channel_Index | Channel_Name | Image
-- | -- | -- | -- | -- | --
img-01.png | dataset01 | 0.0469 | 1 | DAPI | 36638
[...]

PR #67, opened 2022-02-03
Add --detect_header and DatasetColumn support

muhanadz:
## Changes
- This PR adds support to `DatasetColumn`, specifically support to a Dataset ID column

- This PR adds an option CLI `--detect_header` parameter to `omero metatdata populate` that automatically detects `column_type`. This aims to reduce the workload and remove the need for users to predefine the header types through automation.

Dataset ID column was necessary so that the `--detect_header` is easier and better implemented.

Tests passed locally with no failures.

Edit: Changed the default behavior to use the detect header method unless the user supplies a CSV with a custom header with `# header` or the new `--manual_header` flag. 

To do:

- [ ] Update README.md to reflect proper use of `--detect_header`
- [x] Fix other contexts when using `--detect_header`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -127,7 +127,8 @@ setup(
         'future',
         'omero-py>=5.6.0',
         'PyYAML',
-        'jinja2'
+        'jinja2',
+        'pandas'
     ],
     python_requires='>=3',
     tests_require=[
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -245,7 +245,7 @@ class MetadataControl(BaseControl):
             "Allow empty values to become Nan in Long or Double columns"))
 
         populate.add_argument("--manual_header", action="store_true", help=(
-            "Disable automatic header detection row to populate"))
+            "Disable automatic header detection during population"))
 
         populateroi.add_argument(
             "--measurement", type=int, default=None,
@@ -574,7 +574,7 @@ class MetadataControl(BaseControl):
         # Check if first row contains `# header`
         first_row = pd.read_csv(args.file, nrows=1, header=None)
         if not args.manual_header and \
-                not first_row[0].str.contains('# header'):
+                not first_row[0].str.contains('# header').bool():
             omero_metadata.populate.log.info("Detecting header types")
             header_type = self.detect_headers(args.file)
             if args.dry_run:
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -244,8 +244,8 @@ class MetadataControl(BaseControl):
         populate.add_argument("--allow_nan", action="store_true", help=(
             "Allow empty values to become Nan in Long or Double columns"))
 
-        populate.add_argument("--detect_header", action="store_true", help=(
-            "Automatically detect header row to populate"))
+        populate.add_argument("--manual_header", action="store_true", help=(
+            "Disable automatic header detection row to populate"))
 
         populateroi.add_argument(
             "--measurement", type=int, default=None,
@@ -569,10 +569,18 @@ class MetadataControl(BaseControl):
                 md.linkAnnotation(cfgann)
 
         header_type = None
-        if args.detect_header:
+        # To use auto detect header by default unless instructed not to
+        # AND
+        # Check if first row contains `# header`
+        first_row = pd.read_csv(args.file, nrows=1, header=None)
+        if not args.manual_header and \
+                not first_row[0].str.contains('# header'):
+            omero_metadata.populate.log.info("Detecting header types")
             header_type = self.detect_headers(args.file)
             if args.dry_run:
                 omero_metadata.populate.log.info(f"Header Types:{header_type}")
+        else:
+            omero_metadata.populate.log.info("Using user defined header types")
         loops = 0
         ms = 0
         wait = args.wait
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -504,12 +504,15 @@ class MetadataControl(BaseControl):
             col = cols[index]
             if col.lower() in conserved_headers:
                 headers.append(col.lower())
-            elif col.lower() == 'image name' or col.lower() == 'imagename' or \
-                    col.lower() == 'image_name':
+            elif col.lower() == 'image id' or col.lower() == 'imageid' or \
+                    col.lower() == 'image_id':
                 headers.append('image')
-            elif col.lower() == 'dataset name' or \
-                    col.lower() == 'datasetname' or \
-                    col.lower() == 'dataset_name':
+            elif col.lower() == 'roi id' or col.lower() == 'roiid' or \
+                    col.lower() == 'roi_id':
+                headers.append('roi')
+            elif col.lower() == 'dataset id' or \
+                    col.lower() == 'datasetid' or \
+                    col.lower() == 'dataset_id':
                 headers.append('dataset')
             elif col.lower() == 'plate name' or col.lower() == 'platename' or \
                     col.lower() == 'plate_name':
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -32,6 +32,8 @@ from omero.util.metadata_utils import NSBULKANNOTATIONSRAW
 from omero.grid import LongColumn
 from omero.model.enums import UnitsLength
 
+import pandas as pd
+
 HELP = """Metadata utilities
 
 Provides access to and editing of the metadata which
@@ -241,6 +243,9 @@ class MetadataControl(BaseControl):
 
         populate.add_argument("--allow_nan", action="store_true", help=(
             "Allow empty values to become Nan in Long or Double columns"))
+
+        populate.add_argument("--detect_header", action="store_true", help=(
+            "Automatically detect header row to populate"))
 
         populateroi.add_argument(
             "--measurement", type=int, default=None,
@@ -483,6 +488,44 @@ class MetadataControl(BaseControl):
         if not initialized:
             self.ctx.die(100, "Failed to initialize Table")
 
+    def detect_headers(self, csv_path):
+        '''
+        Function to automatically detect headers from a CSV file. This function
+        loads the table to pandas to detects the column type and match headers
+        '''
+
+        conserved_headers = ['well', 'plate', 'image', 'dataset', 'roi']
+        headers = []
+        table = pd.read_csv(csv_path)
+        col_types = table.dtypes.values.tolist()
+        cols = list(table.columns)
+
+        for index, col_type in enumerate(col_types):
+            col = cols[index]
+            if col.lower() in conserved_headers:
+                headers.append(col.lower())
+            elif col.lower() == 'image name' or col.lower() == 'imagename' or \
+            col.lower() == 'image_name':
+                headers.append('image')
+            elif col.lower() == 'dataset name' or col.lower() == 'datasetname' or \
+            col.lower() == 'dataset_name':
+                headers.append('dataset')
+            elif col.lower() == 'plate name' or col.lower() == 'platename' or \
+            col.lower() == 'plate_name':
+                headers.append('plate')
+            elif col.lower() == 'well name' or col.lower() == 'wellname' or \
+            col.lower() == 'well_name':
+                headers.append('well')
+            elif col_type.name == 'object':
+                headers.append('s')
+            elif col_type.name == 'float64':
+                headers.append('d')
+            elif col_type.name == 'int64':
+                headers.append('l')
+            elif col_type.name == 'bool':
+                headers.append('b')
+        return headers
+
     # WRITE
 
     def populate(self, args):
@@ -521,6 +564,12 @@ class MetadataControl(BaseControl):
                 cfgid = cfgann.getFile().getId()
                 md.linkAnnotation(cfgann)
 
+        header_type = None
+        if args.detect_header:
+            header_type = self.detect_headers(args.file)
+            if args.dry_run:
+                omero_metadata.populate.log.info(f"Header Types:{header_type}")
+        # add condition col_type = blarg, open arg.file, arg.detect_header
         loops = 0
         ms = 0
         wait = args.wait
@@ -533,7 +582,7 @@ class MetadataControl(BaseControl):
                             cfg=args.cfg, cfgid=cfgid, attach=args.attach,
                             options=localcfg, batch_size=args.batch,
                             loops=loops, ms=ms, dry_run=args.dry_run,
-                            allow_nan=args.allow_nan)
+                            allow_nan=args.allow_nan, column_types=header_type)
         ctx.parse()
 
     def rois(self, args):
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -505,16 +505,17 @@ class MetadataControl(BaseControl):
             if col.lower() in conserved_headers:
                 headers.append(col.lower())
             elif col.lower() == 'image name' or col.lower() == 'imagename' or \
-            col.lower() == 'image_name':
+                    col.lower() == 'image_name':
                 headers.append('image')
-            elif col.lower() == 'dataset name' or col.lower() == 'datasetname' or \
-            col.lower() == 'dataset_name':
+            elif col.lower() == 'dataset name' or \
+                    col.lower() == 'datasetname' or \
+                    col.lower() == 'dataset_name':
                 headers.append('dataset')
             elif col.lower() == 'plate name' or col.lower() == 'platename' or \
-            col.lower() == 'plate_name':
+                    col.lower() == 'plate_name':
                 headers.append('plate')
             elif col.lower() == 'well name' or col.lower() == 'wellname' or \
-            col.lower() == 'well_name':
+                    col.lower() == 'well_name':
                 headers.append('well')
             elif col_type.name == 'object':
                 headers.append('s')
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -488,7 +488,8 @@ class MetadataControl(BaseControl):
         if not initialized:
             self.ctx.die(100, "Failed to initialize Table")
 
-    def detect_headers(self, csv_path):
+    @staticmethod
+    def detect_headers(csv_path):
         '''
         Function to automatically detect headers from a CSV file. This function
         loads the table to pandas to detects the column type and match headers
@@ -576,7 +577,7 @@ class MetadataControl(BaseControl):
         if not args.manual_header and \
                 not first_row[0].str.contains('# header').bool():
             omero_metadata.populate.log.info("Detecting header types")
-            header_type = self.detect_headers(args.file)
+            header_type = MetadataControl.detect_headers(args.file)
             if args.dry_run:
                 omero_metadata.populate.log.info(f"Header Types:{header_type}")
         else:
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -569,7 +569,6 @@ class MetadataControl(BaseControl):
             header_type = self.detect_headers(args.file)
             if args.dry_run:
                 omero_metadata.populate.log.info(f"Header Types:{header_type}")
-        # add condition col_type = blarg, open arg.file, arg.detect_header
         loops = 0
         ms = 0
         wait = args.wait
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -1556,7 +1556,7 @@ class BulkToMapAnnotationContext(_QueryContext):
     def __init__(self, client, target_object, file=None, fileid=None,
                  cfg=None, cfgid=None, attach=False, options=None,
                  batch_size=1000, loops=10, ms=10, dry_run=False,
-                 allow_nan=False):
+                 allow_nan=False, **kwargs):
         """
         :param client: OMERO client object
         :param target_object: The object to be annotated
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -127,7 +127,8 @@ def read(fname):
         'future',
         'omero-py>=5.6.0',
         'PyYAML',
-        'jinja2'
+        'jinja2',
+        'pandas'
```
sbesson, line 131: At least from the IDR perspective, `pandas` is already a requirement in our server virtual environment so I see no issue with having it as a dependency of this plugin.

I do not think it is an issue for the wider deployment and utilisation of the `omero-metadata` plugin by the community either but I will leave others to comment.

src/omero_metadata/cli.py
```diff
@@ -483,6 +488,45 @@ def testtables(self, args):
         if not initialized:
             self.ctx.die(100, "Failed to initialize Table")
 
+    def detect_headers(self, csv_path):
+        '''
+        Function to automatically detect headers from a CSV file. This function
```
sbesson, line 493: Is this detecting working on the optional line starting with the `# header` and/or the line containing the column names?

src/omero_metadata/cli.py
```diff
@@ -533,7 +582,7 @@ def populate(self, args):
                             cfg=args.cfg, cfgid=cfgid, attach=args.attach,
                             options=localcfg, batch_size=args.batch,
                             loops=loops, ms=ms, dry_run=args.dry_run,
-                            allow_nan=args.allow_nan)
+                            allow_nan=args.allow_nan, column_types=header_type)
```
sbesson, line 585: Similarly to https://github.com/ome/omero-metadata/issues/59, adding a new key-value argument will likely break other contexts e.g. `BulkToMapAnnotationContext`.
 https://github.com/ome/omero-metadata/pull/60 is one way to fix this issue but as indicated in the PR, it would be good to find a more scalable strategy

src/omero_metadata/populate.py
```diff
@@ -825,7 +821,10 @@ def get_image_name_by_id(self, iid, did=None):
 
     def resolve_dataset(self, column, row, value):
         try:
-            return self.datasets_by_name[value].id.val
+            if column.name.lower() == 'dataset':
```
sbesson, line 824: I assume these improvements build on top of the `DatasetColumn` implementation in https://github.com/ome/omero-py/pull/309. Is the requirement for the new API only client-side or will a server need to be upgraded with a version of `omero-py` containing these changes to use this new functionality?

sbesson (commented):
A few questions around new dependencies, client/server requirements as well as the scope of the header detection utility.

I have not carried out any functional testing but as the metadata population is part of the IDR workflow, I suspect we should be able to test it in the context of upcoming submissions. This plugin is also heavily used in the context of our training/guides so this might be another context for testing and providing feedback.

src/omero_metadata/cli.py
muhanadz: It is working on the column names, which removes the requirement for a `# header`. If someone still wants to use the `# header` workflow, they still should be able to.

This reminded me that this script would probably fail with a CSV that has `# header` as a header so perhaps I should add a check whether `# header` is present or not and if it is, then don't use this method even if the flag is present.

src/omero_metadata/populate.py
muhanadz: Correct. During testing, I needed to upgrade my server omero-py with the new DatasetColumn implementation to use the new `--detect_header` functionality. Otherwise, I'd get an exception about DatasetColum not implemented when using a dataset ID column with `'dataset'` as a header type .

src/omero_metadata/cli.py
muhanadz: I totally missed this, I was under the impression that column_types is already a present key-value argument, [here](https://github.com/ome/omero-metadata/blob/148b6a4d4e5cf1deb78e3f087b5907024c749828/src/omero_metadata/populate.py#L989) in the `populate` option, and I'm just now passing in a non-None value instead of the default None.
 
But currently, `omero metadata populate Screen:51 -n --file example.csv --detect_header --context bulkmap` does fail with a similar exception. I will discuss with @emilroz to see if we could brainstorm a more scalable solution.

sbesson: No worries. The absence of integration tests for the other contexts does not help detecting this use case. 
As mentioned in https://github.com/ome/omero-metadata/issues/59#issue-946613849, using `**kwargs` might be an option to handle undefined keywords arguments in a lenient manner.

src/omero_metadata/populate.py
sbesson: Do you remember which call led to the server exception and the exception type thrown by the server? As a general rule, we don't want this plugin to not make particular assumptions. It should be possible to handle both scenarios and fall back to the former logic if `DatasetColumn` is not implemented server-side.  Catching the appropriate exception with a `try/except` block is the immediate solution that comes to mind unless @joshmoore can think of an alternate way to introspect the server-side OMERO.py version.

joshmoore: Other versions are inspectable from the config service, but we'd need to teach the Java process how to inspect the Python process to get that info.

will-moore:
With csv:

```
image,dataset,ROI_Area,Channel_Index,Channel_Name
3659,1451,0.0469,1,DAPI
3686,1451,0.142,2,GFP-AuroraB
3746,1451,0.093,3,TRITC-edit
```

```
$ omero metadata populate --file /Users/wmoore/Desktop/METADATA/test-project-dataset-noheader.csv Project:1001 --detect_header
Created session for will@localhost:4064. Idle timeout: 10 min. Current group: Lab Data
INFO:omero_metadata.populate:Adding dataset:1451 image:3746
INFO:omero_metadata.populate:Adding dataset:1451 image:3686
INFO:omero_metadata.populate:Adding dataset:1451 image:3659
INFO:omero_metadata.populate:Created new table OriginalFile:75983
```

Created a table - viewed with https://github.com/ome/omero-web/pull/355

Screenshot 2022-02-15 at 11 06 45


The column types are:
```
"ImageColumn",
"DatasetColumn",
"DoubleColumn",
"LongColumn",
"StringColumn",
"StringColumn"
```
👍

will-moore:
I noticed in the screenshot above that `dataset` column name isn't capitalised like `image` -> `Image` etc, and other column types as discussed at https://github.com/ome/omero-metadata/pull/56
Would be good to ensure 'Dataset' is the name for any 'DatasetColumn' to be consistent.

src/omero_metadata/populate.py
```diff
@@ -444,7 +440,17 @@ def resolve(self, column, value, row):
...
+                # because the column is image name
+                try:
+                    return images_by_id[int(value)].id.val
+                except ValueError:
+                    val_list = list(images_by_id.values())
+                    key_list = list(images_by_id.keys())
+                    for val in val_list:
+                        if val.name.val == value:
+                            index = val_list.index(val)
+                            return key_list[index]
```
muhanadz, line 453: Found a bug here that when "image name" column is provided on a dataset level, and header type is set to "image", this returns `ValueError` exception since the value passed here is a text-based string and not int. I've fixed this bug however this feels computationally excessive since this fix will search all the values to find the corresponding ID for this text-based string every time. 

Any thoughts are welcome.

muhanadz: @will-moore Understood. Thank you for clarifying. I'll undo the commit/changes.

src/omero_metadata/cli.py
muhanadz: Addressed in 9b158f539f766e9338f3742e5ed0cc68b1769e59
@sbesson Now that `**kwargs` is used for `DeleteMapAnnotationContext` and `BulkToMapAnnotationContext`, there are some key-value pairs that are not used an could be cleaned up. Do you think it's appropriate to clean up here or later in a different PR?

sbesson: If that's okay with you I think cleaning up unused arguments in a follow-up PR would likely simplify the review process

src/omero_metadata/cli.py
sbesson: Re-reading that, I had missed that in addition to detecting the OMERO object types (which I think was previously handled), this also attempts to auto-detect numeric columns. I assume this detection is all delegated to the underlying `pandas` type detection?

It might be worth testing across a few examples but I think this feature would be particular powerful in many of the open-source cases including training and IDR /cc @pwalczysko @will-moore @francesw.
It is pretty clear that the need to specify a ` # header` row is particularly cumbersome (and quite also a non-standard CSV) and thus in many cases, most non OMERO columns end up as simple `StringColumn`. On the other end, there is an ecosystem of tools/analytics that can consume more effectively numeric columns e.g. `omero-parade`, notebooks or more simply Web client table queries.

src/omero_metadata/cli.py
```diff
@@ -242,6 +244,9 @@ def _configure(self, parser):
         populate.add_argument("--allow_nan", action="store_true", help=(
             "Allow empty values to become Nan in Long or Double columns"))
 
+        populate.add_argument("--detect_header", action="store_true", help=(
```
sbesson, line 247: Echoing my enthusiastic comment in https://github.com/ome/omero-metadata/pull/67/files#r816998882, assuming the functionality works as expected, is there any reason not to make this behavior the default and allow to disable the auto-detection via CLI instead?

In terms of the legacy behavior, we might also need to clarify the expectation for users when both the header auto-detection and the `# headers` row are present. Based on my reading of the code, the header auto-detection should set `column_types` and take precedence over the content of the CSV, is that correct?

src/omero_metadata/populate.py
sbesson: Makes sense. In terms of functional review, I think this means we should probably test this tool against two types of servers: with and without upgraded OMERO.py. The population should succeed in both cases but only the server with an upgraded OMERO.py should populate a `DatasetColumn`.
@muhanadz can you add some minimal handling to distinguish between upgraded and non-upgraded servers?

will-moore:
Tested:

```
$ omero metadata populate Project:1851 --file no-header-imagename.csv --detect_header
```
With CSV from README, with no header:
```
Image Name,Dataset Name,ROI_Area,Channel_Index,Channel_Name
img-01.png,dataset01,0.0469,1,DAPI
img-02.png,dataset01,0.142,2,GFP
```
Got column types: `"StringColumn", "StringColumn", "DoubleColumn", "LongColumn", "StringColumn", "ImageColumn"`


With CSV from README, with no header, using `--detect_header`:
```
Well,Plate,Drug,Concentration,Cell_Count,Percent_Mitotic
A1,Index.idx.xml,DMSO,10.1,10,25.4
A2,Index.idx.xml,DMSO,0.1,1000,2.54
```
Got column types (with new cols Well Name & Plate Name)
`"WellColumn","PlateColumn","StringColumn","DoubleColumn","LongColumn","DoubleColumn","StringColumn","StringColumn"`

With CSV from the README, with no header, using `--detect_header` on target Image:
```
Roi,object,probability,area
230422,1,0.8,250
230423,1,0.9,500
```
Got new column "Roi Name" and column types:
"RoiColumn","LongColumn","DoubleColumn","LongColumn","StringColumn"

Looking good.
In all these cases, it seems that the default behaviour should be `--detect_header = True` so that users don't need to remember it every time.

will-moore:
Discussing the `detect_headers` feature with @dominikl 
It would be handy to have this option available directly in the `ParsingContext()` class, which is our entry-point from various python scripts (e.g. https://github.com/IDR/idr0101-payne-insitugenomeseq/blob/560e389dc248976693ae3e647858fdb247c41a72/scripts/csv_to_points.py#L151)

I guess you could do this (not tested yet):
```
from omero_metadata.cli import detect_headers

ParsingContext(....column_types=detect_headers(file))
```
but a `detect_headers` option for `ParsingContext()` would be nicer?
Best of all would be if `detect_headers=True` by default.

sbesson:
> but a detect_headers option for ParsingContext() would be nicer?

Or have some value e.g. `ParsingContext(..., column_types='autodetect')` or similar that would internally trigger the header detection logic?

>  Best of all would be if detect_headers=True by default.

Agreed, that echoes my https://github.com/ome/omero-metadata/pull/67/files#r817003444 and I am still for making this the default assuming the detection is robust (and there is a CLI way to disable it). I still think we need to establish the order of precedence if  header detection is selected and the `# header` row are provided.

sbesson:
See also https://github.com/IDR/idr0062-blin-nuclearsegmentation/pull/12/ for an ongoing IDR annotation investigation using the new logic proposed in this PR as a library

will-moore:
@sbesson I don't see that https://github.com/IDR/idr0062-blin-nuclearsegmentation/pull/12 is using this PR (yet)?

sbesson:
Not formally, but the logic introduced currently as `detect_headers` is very similar to https://github.com/IDR/idr0062-blin-nuclearsegmentation/pull/12/files#diff-4e4de1329ccfc37299b279819006a581e50b7e8a449f2da8c3c009d17063839cR45-R56 so I assume that if this API was migrated to `omero_metadata.library` or even handled internally as discussed above, https://github.com/IDR/idr0062-blin-nuclearsegmentation/pull/12 could be simplified to consume the new API?

will-moore:
I think if a `#header` row is provided on the CSV, then it should take precedence over the auto-detection, since that gives you more control, and reduces the chance of a breaking change.

joshmoore:
> I was never a fan of this custom format

I was going to try to say something to this effect as well. CSV "specifications" are a nightmare, making this feel a bit like the "comments in JSON" conversation.

will-moore:
Sounds good!
So, the only time a user would need the `--manual_header` would be if the csv didn't have `#header` and they wanted e.g. number columns to be treated as strings, or if there was some bug in the auto-detect?
👍

muhanadz:
> Sounds good! So, the only time a user would need the `--manual_header` would be if the csv didn't have `#header` and they wanted e.g. number columns to be treated as strings, or if there was some bug in the auto-detect? 👍

@will-moore: What I had in mind was that `--manual_header` would disable the automatic header detection (currently enabled via `--detect_header` which will be enabled by default). The user would use `--manual_header` when they want to explicitly define the header types via `# header` and pass in a CSV with `# header` as the first row. 

Of course, this would come with some automation, for example, if a user passes in a `# header` CSV without using the `--manual_header` flag, the script would automatically switch to the previous behavior and use `# header`.

will-moore:
Looking to use `detect_headers()` from https://github.com/IDR/idr0062-blin-nuclearsegmentation/pull/12
I didn't want to create an instance of `MetadataControl` class, so it would be helpful to make it a `@staticmethod`:

```
-    def detect_headers(self, csv_path):
+    @staticmethod
+    def detect_headers(csv_path):
```

Then, in that script you can do:
```
from omero_metadata.cli import MetadataControl

column_types = MetadataControl.detect_headers(file_path)
ctx = ParsingContext(..... column_types)
```

will-moore (approved):
Tested again as above without needing any extra arguments this time, and default behaviour created columns of the correct types:
```
$ omero metadata populate Project:801 --file no-header-imagename.csv
```

sbesson (commented):
I did some functional testing the latest version of this PR using a simple P/D/I structure with one dataset and 2 images and a variation of CSV files with/without the header rows and with/without 

```
(venv3) [sbesson@pilot-idr0124-omeroreadwrite ~]$ cat headers_datasetid.csv 
# header dataset,s,s,s
Dataset ID,Image Name,string,float
7459,test.fake,foo,0.1
7459,test2.fake,bar,0.2
(venv3) [sbesson@pilot-idr0124-omeroreadwrite ~]$ cat headers_datasetname.csv 
# header s,s,s,s
Dataset Name,Image Name,string,float
test,test.fake,foo,0.1
test,test2.fake,bar,0.2
(venv3) [sbesson@pilot-idr0124-omeroreadwrite ~]$ cat noheaders_datasetid.csv 
dataset id,image name,string,float
7459,test.fake,foo,0.1
7459,test2.fake,bar,0.2
(venv3) [sbesson@pilot-idr0124-omeroreadwrite ~]$ cat noheaders_datasetname.csv 
Dataset Name,Image Name,string,float
test,test.fake,foo,0.1
test,test2.fake,bar,0.2

```
The behavior of the various implementations (with/without `--manual_header`) matches my expectation from the last commit.

I also briefly tested the scenario discussed in https://github.com/ome/omero-metadata/pull/67/files#r799518087. Trying to create a `DatasetColumn` on a server without the minimal OMERO.py version results in an `omero::InternalException` with `AttributeError: 'DatasetColumn' object has no attribute 'descriptor'`. However, the same error can be thrown without this PR when using `# header dataset,...` so it's better to capture this ass a separate issue to handle this scenario more gracefully.

I spent a bit of time reviewing the different strategies used for choosing the column types. As of the current state of this PR:
- a `# header` row can be stored as the first row of the CSV using the keys defined in https://github.com/ome/omero-metadata/blob/2c1b2691dd61f42ac860ee12095829b1523fed95/src/omero_metadata/populate.py#L120. `HeaderResolver.get_column_types` contains most of the logic for mapping these keys into OMERO.tables columns
- a `column_types` argument can be passed manually  to `ParsingContext` using the same keys as above - see also https://github.com/ome/omero-metadata/issues/6 for a RFE to expose this
- prior to this PR, in the absence of a specified `column_types` and header row, the `HeaderResolver` logic maps column types using column names first and defaulting to StringColumn - https://github.com/ome/omero-metadata/blob/99383af2e50af35f8669aad99dc1f683b2fd67f1/src/omero_metadata/populate.py#L280-L293 contains the relevant logic
- this PR introduces a new logic to detect column types based on column names first and then defaulting to the column type detecting by the `pandas` library. This is currently stored as a `column_types` argument and passed to the `ParsingContext` as above

The last two strategies are  very similar in the intent but there are a few key divergences:
- the `HeaderResolver` detection logic is target dependent i.e. the column names are specific to the target class and defined in the `<class>_keys` variables of the `HeaderResolver` instance. On the other side, the new detection logic is agnostic to the target class i.e. a `well` column will always be mapped as a `WellColumn` even if the population is run on a P/D/I hierarchy
- there is not a perfect overlap between former column names and new column names. Also keys like "plate name" are mapped into `PlateColumn` while e.g. `image_name` is mapped into a `StringColumn` in the current logic

In summary, the new logic is functional and matches the description of the PR. The support for DatasetColumn support and pandas based type detection are very valuable additions.

My outstanding concern is about maintaining and supporting two separate  code paths for the detection of column types using different sets of rules. Several open issues and conversations already point at the unclarity of the behavior of the current OMERO.tables population logic depending on the target type - see https://github.com/ome/omero-metadata/pull/66#issuecomment-1011950209.

Before merging this my preferred option would be to collect and review the different behaviors and unify the name-based column types detection changes. From then we can decide whether we still want to keep the default `StringColumn` mapping or use exclusively the new `pandas` based detection. 

@muhanadz are there some new name-based mappings that are essential to introduce for your workflows?

Happy to schedule a call to review the different options to move forward with this if it makes things easier /cc @pwalczysko

muhanadz:
@sbesson Thank you for the in-depth review and analysis! Some of the issues you raised were definitely a concern of mine. I think a call to review the options would really help. 

The workflow that pushed for the old `--detect_header` implementation was one that couldn't provide `# header` in the csv and the default `StringColumn` wasn't enough. In other words, a user would like to have set headers (eg dataset id) which would automatically be detected as the relevant column type (eg a dataset column in this case) without the `# header` row.

will-moore:
Re Seb `Also keys like "plate name" are mapped into PlateColumn while e.g. image_name is mapped into a StringColumn in the current logic`...
Confusingly, you don't create a csv with `plate_name` but rather with `Plate` column which is changed from string to `PlateColumn` and a new `Plate Name` column gets added.
This is in contrast to the more logical behaviour of `Dataset Name` column which stays as a `StringColumn` although a Dataset column doesn't get added to the table.

sbesson:
I tried to capture my understanding of the different scenarios for the released version of the plugin in https://hackmd.io/EahBQovyR_OyY2PgxewlIg. Feel free to review/amend ahead of the discussion

sbesson:
Summarizing the conversation with @will-moore and @muhanadz. There was an extensive review of the current understanding of the support (https://hackmd.io/EahBQovyR_OyY2PgxewlIg#fn7) as well as the different locations in the code (header resolution, value resolution) which control how the OMERO.tables is populated and the various edge cases/limitations.

Summarizing the discussion:

- overall, there was a global agreement on keeping the detection  target agnostic as proposed in this PR i.e. a `well` column should always map into a `WellColumn` moving forward. Also we want to support the suffixed `<object> id` and `<object> name` columns (with all versions of space, no space, underscore) for all objects
- assuming the new auto-detection rules break some use cases, the manual specification via a `# header` row should allow to manual input the expected behavior
- the new behavior will bring a few usability issues that will need to be collected and triaged
   * what should be the behavior when resolving an object outside the hierarchy e.g. `well` and `Project`. At the moment, the code will error but the feeling is that it could be turned to be lenient. This is also the behavior that was suggested in https://github.com/ome/omero-metadata/issues/51
  * once we have sorted the column mapping, we will likely need to fix the implementation object can be equivalently specified via name and/or ID in all cases - see https://github.com/ome/omero-metadata/issues/65 and https://github.com/ome/omero-metadata/issues/29
  * we might need to improve the code handling and the error message for OMERO.server with older OMERO.py without the `DatasetColumn` implementation

I will merge this PR and let @muhanadz open follow-up PRs with unit tests  as well as documentation updates capturing and describing the new plugin behavior. We can then review this behavior in a wider context e.g. including in various scenarios like IDR or the training sessions for the open-source team, have a general sign-off on the new strategy, deprecate/remove the former code, prioritize the outstanding issues and then move towards a release of the plugin (`0.11.0` or maybe simply `1.0.0`??)

Issue #21, opened 2019-05-29
UnboundLocalError: local variable 'value' referenced before assignment


Borthralla:
I tried running `bin/omero metadata populate --file annotations.csv Project:201` and got the followong output: 
```
Using session for root@localhost:4064. Idle timeout: 10 min. Current group: system
INFO:omero.util.populate_metadata:Adding dataset:505 image:40456
INFO:omero.util.populate_metadata:Adding dataset:505 image:40455
INFO:omero.util.populate_metadata:Adding dataset:505 image:40454
INFO:omero.util.populate_metadata:Adding dataset:505 image:40453
INFO:omero.util.populate_metadata:Adding dataset:505 image:40452
INFO:omero.util.populate_metadata:Adding dataset:505 image:40451
INFO:omero.util.populate_metadata:Adding dataset:505 image:40450
INFO:omero.util.populate_metadata:Adding dataset:505 image:40449
INFO:omero.util.populate_metadata:Adding dataset:505 image:40448
INFO:omero.util.populate_metadata:Adding dataset:505 image:40447
INFO:omero.util.populate_metadata:Created new table OriginalFile:54420
Traceback (most recent call last):
  File "bin/omero", line 130, in <module>
    rv = omero.cli.argv()
  File "/home/omero/OMERO.server-5.4.9-ice36-b101/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "/home/omero/OMERO.server-5.4.9-ice36-b101/lib/python/omero/cli.py", line 1095, in invoke
    stop = self.onecmd(line, previous_args)
  File "/home/omero/OMERO.server-5.4.9-ice36-b101/lib/python/omero/cli.py", line 1172, in onecmd
    self.execute(line, previous_args)
  File "/home/omero/OMERO.server-5.4.9-ice36-b101/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "/usr/local/lib/python2.7/dist-packages/omero_cli_metadata.py", line 491, in populate
    ctx.parse()
  File "/usr/local/lib/python2.7/dist-packages/populate_metadata.py", line 993, in parse
    return self.parse_from_handle_stream(data)
  File "/usr/local/lib/python2.7/dist-packages/populate_metadata.py", line 963, in parse_from_handle_stream
    self.populate_from_reader(reader, self.filter_function, table, 1000)
[...]

PR #70, opened 2022-04-06
Skip empty rows

dominikl:
Simply skip empty lines instead of crashing. Ran into this issue as well. Initially thought it's a linebreak at the end of the last entry, but that works fine actually. In order to replicate the issue you really need an empty line with a linebreak at the end of the file. As a side effect with the PR you can also have empty lines somewhere inbetween, they too will just be ignored.
Fixes https://github.com/ome/omero-metadata/issues/21 .

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -1314,7 +1314,7 @@ class ParsingContext(object):
                             column.values = []
                         row_count = 0
                 else:
-                    log.info('Skip empty row %d', r + 1)
+                    log.warning('Skip empty row %d', r + 1)
         if row_count != 0:
             log.debug("DATA TO ADD")
             log.debug(self.columns)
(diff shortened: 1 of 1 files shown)


src/omero_metadata/populate.py
```diff
@@ -1304,14 +1304,17 @@ def populate_from_reader(self,
...
+                    self.populate_row(row)
+                    row_count = row_count + 1
+                    if row_count >= batch_size:
+                        self.post_process()
+                        table.addData(self.columns)
+                        for column in self.columns:
+                            column.values = []
+                        row_count = 0
+                else:
+                    log.info('Skip empty row %d', r + 1)
```
sbesson, line 1317: I would elevate this warning and the one below to `WARN` to make sure the presence of such lines does not get silently handled

will-moore:
Tested connecting to merge-ci and it works fine. 

```
$ omero metadata populate Project:857 --file ../csv-tests/project\:857.csv
INFO:omero_metadata.populate:Created new table OriginalFile:1280916
WARNING:omero_metadata.populate:Skip empty row 3
```

sbesson (approved):
As #67 has been integrated but we are still waiting on documentation changes to agree on the behavior of the next `omero-metadata` release, the `master` branch is currently not releasable.

Does this functionality need to be released immediately or can it wait and be integrated together with the header detection API? @dominikl

dominikl:
I think it can wait, not urgent.

PR #71, opened 2022-04-07
Updating README

muhanadz:
Updating README.rst to include the changes to the default behavior from https://github.com/ome/omero-metadata/pull/67.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,7 +72,7 @@ The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``
 
 OMERO.tables have defined column types to specify the data-type such as `double` or `long` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
 
-**The default behaviour of the script is to automatically detect these column types from an input ``CSV``**. This behaviour works as folows:
+The default behaviour of the script is to ``automatically detect these column types`` from an input ``CSV``. This behaviour works as folows:
 
 *  Columns named with a supported object-type ('plate', 'well', 'image', 'dataset', or 'roi') or with object_id (e.g. 'image_id', 'dataset_id') will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
 *  Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
--- a/README.rst
+++ b/README.rst
@@ -73,8 +73,9 @@ The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``
 OMERO.tables have defined column types to specify the data-type such as `double` or `long` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
 
 **The default behaviour of the script is to automatically detect these column types from an input ``CSV``**. This behaviour works as folows:
-* Columns named with a supported object-type (e.g. 'image', 'plate'...) or with object_id (e.g. 'image_id', 'dataset_id' ) will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
-* Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
+
+*  Columns named with a supported object-type (e.g. 'image', 'plate'...) or with object_id (e.g. 'image_id', 'dataset_id' ) will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
+*  Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
 
 
 However, it is possible to manually define the column types , ignoring the automatic header detection, if a ``CSV`` with a ``# header`` row is passed. The ``# header`` row should be the first row of the CSV and defines columns according to the following list (see examples below):
--- a/README.rst
+++ b/README.rst
@@ -76,25 +76,25 @@ The default behaviour of the script is to automatically detect the column types 
 
 *  Columns named with a supported object-type (e.g. ``plate``, ``well``, ``image``, ``dataset``, or ``roi``), with ``<object> id`` or ``<object> name`` will generate the corresponding column type in the OMERO.table. See table below for full list of supported column names.
 
-============ ================= ==================== ==================================
+============ ================= ==================== ====================================================================
 Column Name  Column type       Detected Header Type Notes
-============ ================= ==================== ==================================
-Image        ``ImageColumn``   ``image``            Appends 'Image Name' column
-Image Name   ``StringColumn``  ``s``                Appends 'Image' column
-Image ID     ``ImageColumn``   ``image``            Appends 'Image Name' column
-Dataset      ``DatasetColumn`` ``dataset``          \-
-Dataset Name ``StringColumn``  ``s``                \-
-Dataset ID   ``DatasetColumn`` ``dataset``          \-
-Plate        ``PlateColumn``   ``plate``            Adds 'Plate' column
-Plate Name   ``PlateColumn``   ``plate``            Adds 'Plate' column
-Plate ID     ``LongColumn``    ``l``                \-
-Well         ``WellColumn``    ``well``             Adds 'Well' column
-Well Name    ``WellColumn``    ``well``             Adds 'Well' column
-Well ID      ``LongColumn``    ``l``                \-
-ROI          ``RoiColumn``     ``roi``              Appends 'ROI Name' column
-ROI Name     ``StringColumn``  ``s``                Appends 'ROI' column
-ROI ID       ``RoiColumn``     ``roi``              Appends 'ROI Name' column
-============ ================= ==================== ==================================
+============ ================= ==================== ====================================================================
+Image        ``ImageColumn``   ``image``            Accepts image IDs. Appends new 'Image Name' column with image names.
+Image Name   ``StringColumn``  ``s``                Accepts image names. Appends new 'Image' column with image IDs.
+Image ID     ``ImageColumn``   ``image``            Accepts image IDs. Appends new 'Image Name' column with image names.
+Dataset      ``DatasetColumn`` ``dataset``          Accepts dataset IDs.
+Dataset Name ``StringColumn``  ``s``                Accepts dataset names.
+Dataset ID   ``DatasetColumn`` ``dataset``          Accepts dataset IDs.
+Plate        ``PlateColumn``   ``plate``            Accepts plate names. Adds new 'Plate' column with plate IDs.
+Plate Name   ``PlateColumn``   ``plate``            Accepts plate names. Adds new 'Plate' column with plate IDs.
+Plate ID     ``LongColumn``    ``l``                Accepts plate IDs.
+Well         ``WellColumn``    ``well``             Accepts well names. Adds new 'Well' column with well IDs.
+Well Name    ``WellColumn``    ``well``             Accepts well names. Adds new 'Well' column with well IDs.
+Well ID      ``LongColumn``    ``l``                Accepts well IDs.
+ROI          ``RoiColumn``     ``roi``              Accepts ROI IDs. Appends new 'ROI Name' column with ROI names.
+ROI Name     ``StringColumn``  ``s``                Accepts ROI names. Appends new 'ROI' column with ROI IDs.
+ROI ID       ``RoiColumn``     ``roi``              Accepts ROI IDs. Appends new 'ROI Name' column with ROI names.
+============ ================= ==================== ====================================================================
          
 Note: Column names are case insensitive. Space, no space, and underscore are all accepted as separators for column names (i.e. ``<object> name``/``<object> id```, ``<object>name``/``<object>id``, ``<object>_name``/``<object>_id`` are all accepted)
 
--- a/README.rst
+++ b/README.rst
@@ -70,9 +70,9 @@ object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-OMERO.tables have defined column types to specify the data-type such as `double` or `long` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
+OMERO.tables have defined column types to specify the data-type such as ``double`` or ``long`` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
 
-The default behaviour of the script is to ``automatically detect these column types`` from an input ``CSV``. This behaviour works as folows:
+The default behaviour of the script is to automatically detect the column types from an input ``CSV``. This behaviour works as folows:
 
 *  Columns named with a supported object-type ('plate', 'well', 'image', 'dataset', or 'roi') or with object_id (e.g. 'image_id', 'dataset_id') will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
 *  Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
--- a/README.rst
+++ b/README.rst
@@ -97,7 +97,7 @@ Examples
 **Project / Dataset**
 ^^^^^^^^^^^^^^^^^^^^^^
 
-To add a table to a Project, the ``CSV`` file needs to specify ``Dataset Name``
+To add a table to a Project, the ``CSV`` file needs to specify ``Dataset Name`` or ``Dataset ID``
 and ``Image Name`` or ``Image ID``::
 
     $ omero metadata populate Project:1 --file path/to/project.csv
@@ -131,8 +131,38 @@ img-03.png dataset01    0.093    3             TRITC        36640
 img-04.png dataset01    0.429    4             Cy5          36641
 ========== ============ ======== ============= ============ =====
 
-If the target is a Dataset instead of a Project, the ``Dataset Name`` column is not needed.
+Example using ``Image ID`` and ``Dataset ID``.
 
+project.csv (manual column types definition)::
+
+    # header image,dataset,d,l,s
+    image id,Dataset ID,ROI_Area,Channel_Index,Channel_Name
+    36638,101,0.0469,1,DAPI
+    36639,101,0.142,2,GFP
+    36640,101,0.093,3,TRITC
+    36641,101,0.429,4,Cy5
+    
+project.csv (automatic column types detection)::
+
+    image id,Dataset ID,ROI_Area,Channel_Index,Channel_Name
+    36638,101,0.0469,1,DAPI
+    36639,101,0.142,2,GFP
+    36640,101,0.093,3,TRITC
+    36641,101,0.429,4,Cy5
+    
+Both manual definition or automatic detection of column types will create an OMERO.table linked to the Project as folows with
+a new ``Image`` column with Names:
+
+===== ======= ======== ============= ============ ==========
+Image Dataset ROI_Area Channel_Index Channel_Name Image Name
+===== ======= ======== ============= ============ ==========
+36638 101     0.0469   1             DAPI         img-01.png 
+36639 101     0.142    2             GFP          img-02.png 
+36640 101     0.093    3             TRITC        img-03.png 
+36641 101     0.429    4             Cy5          img-04.png
+===== ======= ======== ============= ============ ==========
+
+If the target is a Dataset instead of a Project, the ``Dataset Name`` column is not needed.
 
 **Screen / Plate**
 ^^^^^^^^^^^^^^^^^^^
--- a/README.rst
+++ b/README.rst
@@ -231,11 +231,7 @@ If there is an ``roi`` column (header type ``roi``) containing ROI IDs, an ``Roi
 column will be appended automatically (see example below). If a column of Shape IDs named ``shape``
 of type ``l`` is included, the Shape IDs will be validated (and set to -1 if invalid).
 Also if an ``image`` column of Image IDs is included, an ``Image Name`` column will be added.
-NB: Columns of type ``shape`` aren't yet supported on the OMERO.server.
-
-Alternatively, if the target is an Image, the ROI input column can be
-``Roi Name`` (with type ``s``), and an ``roi`` type column will be appended containing ROI IDs.
-In this case, it is required that ROIs on the Image in OMERO have the ``Name`` attribute set::
+NB: Columns of type ``shape`` aren't yet supported on the OMERO.server::
 
     $ omero metadata populate Image:1 --file path/to/image.csv
 
@@ -262,6 +258,10 @@ Roi shape object probability area Roi Name
 === ===== ====== =========== ==== ========
 
 Note: equivalent to adding ``# header roi,l,l,d,l`` row to the top of the ``image.csv`` for manual definition.
+
+Alternatively, if the target is an Image, the ROI input column can be
+``Roi Name`` (with type ``s``), and an ``roi`` type column will be appended containing ROI IDs.
+In this case, it is required that ROIs on the Image in OMERO have the ``Name`` attribute set.
 
 Note that the ROI-level data from an ``OMERO.table`` is not visible
 in the OMERO.web UI right-hand panel under the ``Tables`` tab,
--- a/README.rst
+++ b/README.rst
@@ -70,7 +70,7 @@ object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-OMERO.tables have defined column types to specify the data-type such as ``double`` or ``long`` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
+OMERO.tables have defined column types to specify the data-type such as ``double`` or ``long`` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``.
 
 The default behaviour of the script is to automatically detect the column types from an input ``CSV``. This behaviour works as follows:
 
--- a/README.rst
+++ b/README.rst
@@ -70,8 +70,12 @@ object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-If you wish to ensure that ``number`` columns are created for numerical data, this will
-allow you to make numerical queries on the table.
+
+
+Automatic header
+^^^^^^^^^
+
+**The default behaviour of the script is to automatically detect the column types and specific object types from an input ``CSV`` using the list below.**
 Column Types are:
 
 - ``d``: ``DoubleColumn``, for floating point numbers
@@ -80,8 +84,14 @@ Column Types are:
 - ``b``: ``BoolColumn``, for true/false
 - ``plate``, ``well``, ``image``, ``dataset``, ``roi`` to specify objects
 
-These can be specified in the first row of a ``CSV`` with a ``# header`` tag (see examples below).
-The ``# header`` row is optional. Default column type is ``String``.
+
+
+Manual Header
+^^^^^^^^^
+
+However, it is possible to override the default behaviour, ignoring the automatic header detection, and manually assign the header to define the column type if a ``CSV`` with with a ``# header`` tag is passed (see examples below).
+
+Automatic header detection can also be ignored if using the ``--manual_headers`` flag. If the ``# header`` is not present and this flag is used, column types will default to ``String``
 
 NB: Column names should not contain spaces if you want to be able to query
 by these columns.
--- a/README.rst
+++ b/README.rst
@@ -70,18 +70,19 @@ object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
+OMERO.tables have defined column types to specify the data-type such as `double` or `long` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
 
-**The default behaviour of the script is to automatically detect the column types and specific object types from an input ``CSV`` using the list below.**
-Column Types are:
+**The default behaviour of the script is to automatically detect these column types from an input ``CSV``**. This behaviour works as folows:
+* Columns named with a supported object-type (e.g. 'image', 'plate'...) or with object_id (e.g. 'image_id', 'dataset_id' ) will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
+* Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
 
+
+However, it is possible to manually define the column types , ignoring the automatic header detection, if a ``CSV`` with a ``# header`` row is passed. The ``# header`` row should be the first row of the CSV and defines columns according to the following list (see examples below):
 - ``d``: ``DoubleColumn``, for floating point numbers
 - ``l``: ``LongColumn``, for integer numbers
 - ``s``: ``StringColumn``, for text
 - ``b``: ``BoolColumn``, for true/false
 - ``plate``, ``well``, ``image``, ``dataset``, ``roi`` to specify objects
-
-
-However, it is possible to manually define the column types , ignoring the automatic header detection, if a ``CSV`` with a ``# header`` row is passed (see examples below).
 
 Automatic header detection can also be ignored if using the ``--manual_headers`` flag. If the ``# header`` is not present and this flag is used, column types will default to ``String`` (unless the column names correspond to OMERO objects such as ``image`` or ``plate``).
 
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ OMERO.tables have defined column types to specify the data-type such as `double`
 
 **The default behaviour of the script is to automatically detect these column types from an input ``CSV``**. This behaviour works as folows:
 
-*  Columns named with a supported object-type (e.g. 'image', 'plate'...) or with object_id (e.g. 'image_id', 'dataset_id' ) will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
+*  Columns named with a supported object-type ('plate', 'well', 'image', 'dataset', or 'roi') or with object_id (e.g. 'image_id', 'dataset_id') will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -80,8 +84,14 @@ Column Types are:
...
 - ``plate``, ``well``, ``image``, ``dataset``, ``roi`` to specify objects
 
-These can be specified in the first row of a ``CSV`` with a ``# header`` tag (see examples below).
-The ``# header`` row is optional. Default column type is ``String``.
+
+
+Manual Column Types
+^^^^^^^^^
+
+However, it is possible to override the default behaviour, ignoring the automatic header detection, and manually assign the header to define the column type if a ``CSV`` with with a ``# header`` tag is passed (see examples below).
```
will-moore, line 92: "with with"
I would provide a single example of the `# header` usage, and remove headers from other CSV examples to show them as we expect users to use them now?
I'm not sure "tag" makes most sense here. You really mean a "row"?

README.rst
```diff
@@ -80,8 +84,14 @@ Column Types are:
...
+
+Automatic header detection can also be ignored if using the ``--manual_headers`` flag. If the ``# header`` is not present and this flag is used, column types will default to ``String``
```
will-moore, line 94: ...unless the column names correspond to OMERO objects such as "image" etc?

muhanadz: Correct. I didn't want to overcomplicate the explanation since not many people will be using this path. Will add more details in brackets.

README.rst
muhanadz: Thanks for catching out the mistake. And yes "tag" is probably inaccurate, will use a "row". I will also fix the examples accordingly

README.rst
muhanadz: Addressed in [0d291cc](https://github.com/ome/omero-metadata/pull/71/commits/0d291cc2263e0f5145c6a82c64b902b4bb13047e)

README.rst
muhanadz: Addressed in [0d291cc](https://github.com/ome/omero-metadata/pull/71/commits/0d291cc2263e0f5145c6a82c64b902b4bb13047e)

README.rst
```diff
@@ -71,10 +71,6 @@ object IDs in the ``OMERO.table``.
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
 
-
-Automatic Column Types
-^^^^^^^^^
-
 **The default behaviour of the script is to automatically detect the column types and specific object types from an input ``CSV`` using the list below.**
```
will-moore, line 74: Reading from the top, it feels like this needs an extra sentence of intro *before* "The default behaviour..."

"OMERO.tables have defined column types to specify the data-type of each column, and special object types for storing OMERO object IDs, such as `ImageColumn` or `WellColumn`"

Then, instead of "...using the list below", I think it needs a summary of the logic.
Something like this:

 - csv columns named with a supported object type ('image', 'plate'...) or with `object_id` (e.g. `image_id`) will generate the corresponding column types in the OMERO.table. e.g `ImageColumn`
 - other column types will be detected based on the data in the columns using the `pandas` library.

Then say "However, it is possible.... 
"The `# header` should be the first row of the CSV and defines columns according to the following list:
[ then have the list of column-types ]

muhanadz: This is really helpful. Thank you! I made these changes with minor changes of mine in [494314a](https://github.com/ome/omero-metadata/pull/71/commits/494314a04b5b1f38eb3076b03edd740fd99a5423). Please let me know if this isn't consistent with the meaning you had in mind.

README.rst
```diff
@@ -70,30 +70,37 @@ object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-If you wish to ensure that ``number`` columns are created for numerical data, this will
-allow you to make numerical queries on the table.
-Column Types are:
+OMERO.tables have defined column types to specify the data-type such as `double` or `long` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
 
+**The default behaviour of the script is to automatically detect these column types from an input ``CSV``**. This behaviour works as folows:
+* Columns named with a supported object-type (e.g. 'image', 'plate'...) or with object_id (e.g. 'image_id', 'dataset_id' ) will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
```
sbesson, line 76: Since we support only a handful of primary objects, would it be worth being explicit and have the full list here e.g. `supported object-type ('plate', 'well', 'image', 'dataset', 'roi')`

README.rst
```diff
@@ -70,30 +70,37 @@ object IDs in the ``OMERO.table``.
...
+* Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
```
sbesson, line 77: The only exception is that column named as `<object> name` will be handled specially right?

will-moore: Yes. Could add a point before the last to explain that. But since that's different for Plate vv Dataset & Image, we need to refer to the examples below.

README.rst
will-moore: Reads fine. Formatting at https://github.com/ome/omero-metadata/blob/494314a04b5b1f38eb3076b03edd740fd99a5423/README.rst is not working for those bullet points.

README.rst
muhanadz: Agreed, addressed in [2bc1a06](https://github.com/ome/omero-metadata/pull/71/commits/2bc1a06aafd76f143d0c246e70ba940720e0df5e)

README.rst
muhanadz: Fixed in [70a1c92](https://github.com/ome/omero-metadata/pull/71/commits/70a1c92a795759342cf4f98fe9d6c03ba85a0a19)

README.rst
muhanadz: Agreed with Will, I think once we are more consistent with all the objects (eg plate_name and plate_id), it'll be easier to explain that here.

README.rst
```diff
@@ -70,30 +70,38 @@ object IDs in the ``OMERO.table``.
...
-Column Types are:
+OMERO.tables have defined column types to specify the data-type such as `double` or `long` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
 
+**The default behaviour of the script is to automatically detect these column types from an input ``CSV``**. This behaviour works as folows:
+
+*  Columns named with a supported object-type ('plate', 'well', 'image', 'dataset', or 'roi') or with object_id (e.g. 'image_id', 'dataset_id') will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
+*  Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
+
+
+However, it is possible to manually define the column types , ignoring the automatic header detection, if a ``CSV`` with a ``# header`` row is passed. The ``# header`` row should be the first row of the CSV and defines columns according to the following list (see examples below):
```
sbesson, line 81: I think this requires an extra empty line to render properly

sbesson (changes requested):
A few remaining rst linting errors to fix.  Two outstanding questions:

- from the UoD team, are there any objections to adopting the new behavior both in the context of IDR as well as the training? /cc @dominikl @pwalczysko 
- should we get the new feature released as `0.10.0` or is it time to call this `1.0.0`? If the latter was desirable but we felt like we need additional evaluation, an intermediate approach would be to get this as a pre-releases e.g. `1.0.0a1`.

README.rst
muhanadz: updated in [f6610a2](https://github.com/ome/omero-metadata/pull/71/commits/f6610a20bbb6914c5843474f2e80d46b9bf8c92d). Sorry haven't worked with rst before, thanks for spotting it!

README.rst
```diff
@@ -70,30 +70,39 @@ object IDs in the ``OMERO.table``.
...
+Automatic header detection can also be ignored if using the ``--manual_headers`` flag. If the ``# header`` is not present and this flag is used, column types will default to ``String`` (unless the column names correspond to OMERO objects such as ``image`` or ``plate``).
 
 NB: Column names should not contain spaces if you want to be able to query
 by these columns.
 
+Examples
+^^^^^^^^^
+
 **Project / Dataset**
+^^^^^^^^^
```
sbesson, line 98:
```
WARNING README.rst:98 Title underline too short.
```

I think we need the lenght of the underline to be at least as long as the heading

README.rst
```diff
@@ -118,11 +135,14 @@ If the target is a Dataset instead of a Project, the ``Dataset Name`` column is
 
 
 **Screen / Plate**
+^^^^^^^^^
```
sbesson, line 138: Same as above

sbesson (changes requested):

sbesson:
Thanks @muhanadz. This is now staged for review at https://github.com/ome/omero-metadata/blob/0cc977f9c4d2309b8712f2ef20ad93cba68295f2/README.rst. As mentioned above, next step is to get a sign off from the team and decide on the versioning/roadmap.

README.rst
```diff
@@ -70,30 +70,39 @@ object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-If you wish to ensure that ``number`` columns are created for numerical data, this will
-allow you to make numerical queries on the table.
-Column Types are:
+OMERO.tables have defined column types to specify the data-type such as `double` or `long` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
+
+**The default behaviour of the script is to automatically detect these column types from an input ``CSV``**. This behaviour works as folows:
```
pwalczysko, line 75:
```suggestion
The default behaviour of the script is to ``automatically detect these column types`` from an input ``CSV``. This behaviour works as folows:
```

I do not see a pressing need for this sentence to be in bold, I think the double star is somehow invalidating the formatting of the ``CSV``, see below.
Screenshot 2022-04-15 at 16 33 05

pwalczysko: @muhanadz More importantly, when reading this sentence, it is not immediately clear what is meant by ``these column types``. In the previous sentence two distinct classes of column types are mentioned (data-type and special object-type). It would be better to formulate the sentence saying for example "both data-type and object-type are atumatically..." or somesuch.

README.rst
```diff
@@ -70,30 +70,39 @@ object IDs in the ``OMERO.table``.
...
+
+*  Columns named with a supported object-type ('plate', 'well', 'image', 'dataset', or 'roi') or with object_id (e.g. 'image_id', 'dataset_id') will generate the corresponding column type in the OMERO.table. e.g (ImageColumn, PlateColumn, DatasetColumn, etc)
```
pwalczysko, line 77: I understand that a column named with a supported object-type ``image`` will generate corresponding column type in the OMERO.table ``ImageColumn`` etc. But what column type is generated when the column is named with ``image_id`` ? I am only guessing that also an a column type ImageColumn is created, but this might not be correct...

README.rst
```diff
@@ -70,30 +70,39 @@ object IDs in the ``OMERO.table``.
...
+*  Other column types will be detected based on the column's data using the pandas library (e.g. column of data type double).
```
pwalczysko, line 78: ``Other column types``
You call them ``data-type`` columns above (line 73) (if I get it right), why not to stick to the same nomenclature here ?

pwalczysko: Or maybe it wants to say "All other columns..." ? Meaning no column will be left undetected ?

README.rst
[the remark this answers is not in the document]
pwalczysko: So the sentence is clearly wrong after running following example (which I made according to my understanding of what the sentence is sayin), I have an error below.

CSV used
```
image_id,Dataset Name,ROI_Area,Channel_Index,Channel_Name
94770,dataset01,0.0469,1,DAPI
94772,dataset01,0.142,2,GFP
94775,dataset01,0.093,3,TRITC
94776,dataset01,0.429,4,Cy5
```

Error

```
omero metadata populate Project:2272 --file  ~/populate-no-roi.csv 
Using session for trainer-1@outreach.openmicroscopy.org:4064. Idle timeout: 10 min. Current group: Lab1
INFO:omero_metadata.populate:Detecting header types
INFO:omero_metadata.populate:Adding dataset:16366 image:95076
INFO:omero_metadata.populate:Adding dataset:16366 image:95075
INFO:omero_metadata.populate:Adding dataset:16366 image:94772
INFO:omero_metadata.populate:Adding dataset:16366 image:94770
  File "/Users/pwalczysko/miniconda3/envs/cli/bin/omero", line 11, in <module>
    sys.exit(main())
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/main.py", line 125, in main
    rv = omero.cli.argv()
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1784, in argv
    cli.invoke(args[1:])
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1222, in invoke
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1299, in onecmd
  File "/Users/pwalczysko/miniconda3/envs/cli/lib/python3.6/site-packages/omero/cli.py", line 1381, in execute
    args.func(args)
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/cli.py", line 598, in populate
    ctx.parse()
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 1216, in parse
    self.preprocess_from_handle(f1)
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 1158, in preprocess_from_handle
    self.preprocess_data(reader)
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 1260, in preprocess_data
    self.post_process()
  File "/Users/pwalczysko/Work/omero-metadata/src/omero_metadata/populate.py", line 1438, in post_process
    dname = columns_by_name["dataset name"].values[i]
IndexError: list index out of range
```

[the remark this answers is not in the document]
pwalczysko: @muhanadz great, thanks, please have a look at the error https://github.com/ome/omero-metadata/pull/71/commits/bf24a6b7c1ffabe231e9da83c0e3e0f5b30f35eb#r852191963  though too and comment please. Surely better examples will be more useful than the sentence, however well formulated. Sorry for being so meticulous here, but I am trying to understand the details here in order to be able to correct the corresponding text in omero-guides.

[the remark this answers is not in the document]
sbesson: Is this possibly related to https://github.com/ome/omero-metadata/issues/65 i.e. specifying `image` or `image id` is currently broken in a P/D/I hierarchy?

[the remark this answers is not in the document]
pwalczysko: > Is this possibly related to #65 i.e. specifying `image` or `image id` is currently broken in a P/D/I hierarchy?

@sbesson thank you very much, this seems like the same problem. The error is the same, and when I tried to leave out the Dataset column and target a Dataset instead of the Project, I was successful just as described in #65

pwalczysko:
Tested the examples listed for the automatic column detection (Project, Dataset, Screen and Image (adding roi table)). Worked fine. Happy to test the new @muhanadz 's examples too.

README.rst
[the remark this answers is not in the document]
muhanadz: I have added an example for a CSV with `dataset id` and `image id` in [6c5500b](https://github.com/ome/omero-metadata/pull/71/commits/6c5500b60b648e3f3f7de86e01af53d18df22bb5). I think this should cover most common scenarios but happy to add more if necessary.

README.rst
```diff
@@ -70,30 +70,39 @@ object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-If you wish to ensure that ``number`` columns are created for numerical data, this will
-allow you to make numerical queries on the table.
-Column Types are:
+OMERO.tables have defined column types to specify the data-type such as ``double`` or ``long`` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
+
+The default behaviour of the script is to automatically detect the column types from an input ``CSV``. This behaviour works as folows:
```
will-moore, line 75: Typo `folows`

README.rst
```diff
@@ -70,30 +70,39 @@ object IDs in the ``OMERO.table``.
...
+
+*  Columns named with a supported object-type (e.g. 'plate', 'well', 'image', 'dataset', or 'roi'), with <object>_id (e.g. 'image_id', 'dataset_id') or with <object>_name (e.g. 'plate_name', 'dataset_name') will generate the corresponding column type in the OMERO.table (e.g. ImageColumn, PlateColumn, DatasetColumn, etc).
```
will-moore, line 77: It's slightly nicer to use the backticks ``plate`` etc for these identifiers. Also ``<object>_id`` and ``<object>_name``, ``ImageColumn`` etc.

This sentence is a little confusing as it suggests that `plate` and `plate_name` will behave in the same way as each other and "generate a PlateColumn" (same for `dataset` and `dataset_name`). But I think it's more complex:
 - `dataset` -> `DatasetColumn`
 -  `dataset_name` (or `Dataset Name` in eg below) -> StringColumn (no `DatasetColumn` added in eg below)
 -  `plate` -> `PlateColumn` (and a new `StringColumn` "Plate Name" gets added)
 - `plate_name`?? - no examples
 - `Image Name` (and `Image_Name`?) -> String column (and ImageColumn added)
 - `image` or `image_id` -> ImageColumn

README.rst
```diff
@@ -102,7 +111,15 @@ project.csv::
...
-This will create an OMERO.table linked to the Project like this with
+project.csv (automatic column types detection)::
+
+    Image Name,Dataset Name,ROI_Area,Channel_Index,Channel_Name
+    img-01.png,dataset01,0.0469,1,DAPI
+    img-02.png,dataset01,0.142,2,GFP
+    img-03.png,dataset01,0.093,3,TRITC
+    img-04.png,dataset01,0.429,4,Cy5
+
+Both manual definition or automatic detection of column types will create an OMERO.table linked to the Project as folows with
```
will-moore, line 122: folows again (and below)

README.rst
will-moore: Need to mention case insensitivity? What about underscore handling? E.g. `image_name` == `Image Name`?

will-moore (changes requested):
Looks good. Nice to have the extra examples.
We haven't done this before, but I think it would be useful if the output tables in the Doc could somehow indicate what type the columns are? E.g. DoubleColumn, ImageColumn etc. 

Whenever we're adding / updating the examples or changing the code, it would be nice to have a way to automatically check that the examples do actually generate the documented output. But not for this PR!

will-moore (commented):
Table looks great. Few comments/queries added...

README.rst
```diff
@@ -64,21 +64,53 @@ populate
...
+============ ================= ==================== ==================================
+Image        ``ImageColumn``   ``image``            Appends 'Image Name' column
+Image Name   ``StringColumn``  ``s``                Appends 'Image' column
+Image ID     ``ImageColumn``   ``image``            Appends 'Image Name' column
+Dataset      ``DatasetColumn`` ``dataset``          \-
+Dataset Name ``StringColumn``  ``s``                \-
+Dataset ID   ``DatasetColumn`` ``dataset``          \-
+Plate        ``PlateColumn``   ``plate``            Adds 'Plate' column
+Plate Name   ``PlateColumn``   ``plate``            Adds 'Plate' column
+Plate ID     ``LongColumn``    ``l``                \-
```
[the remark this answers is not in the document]
muhanadz: According to my tests, `Plate ID`/`Well ID` never gets resolved as a Plate/Well ID and instead of a `LongColumn`. The following table:
```
Plate ID,Well ID,ROI_Area,Channel_Index,Channel_Name
51,61,0.0469,1,DAPI
51,58,0.0469,1,DAPI
51,62,0.0469,1,DAPI
51,55,0.0469,1,DAPI
```
Header detected as `[l,l,d,l,s]`
Output:
image

If you try to use `#header plate,well,d,l,s`, the code fails with `Cannot parse well identifier "Well ID" from row:`. This is definitely something to develop.

There is a difference between `Append` and `Adds`. Appends simply appends column to the end of the table while Adds replaces the original column with its respective column (`Plate Name` becomes `Plate`) and appends the original `Plate Name` column to the end of the table. I made the distinction for us to understand and not necessarily the user to know about.

README.rst
```diff
@@ -64,21 +64,53 @@ populate
...
+Well         ``WellColumn``    ``well``             Adds 'Well' column
+Well Name    ``WellColumn``    ``well``             Adds 'Well' column
+Well ID      ``LongColumn``    ``l``                \-
+ROI          ``RoiColumn``     ``roi``              Appends 'ROI Name' column
+ROI Name     ``StringColumn``  ``s``                \-
```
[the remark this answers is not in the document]
muhanadz: I don't think so, but I will triple-check this.

README.rst
```diff
@@ -64,21 +64,53 @@ populate
...
-However, it is possible to manually define the column types , ignoring the automatic header detection, if a ``CSV`` with a ``# header`` row is passed. The ``# header`` row should be the first row of the CSV and defines columns according to the following list (see examples below):
+The default behaviour of the script is to automatically detect the column types from an input ``CSV``. This behaviour works as follows:
+
+*  Columns named with a supported object-type (e.g. ``plate``, ``well``, ``image``, ``dataset``, or ``roi``), with ``<object>_id`` or ``<object>_name`` will generate the corresponding column type in the OMERO.table. See table below for full list of supported column names.
+
+============ ================= ==================== ==================================
+Column Name  Column type       Detected Header Type Notes
+============ ================= ==================== ==================================
+Image        ``ImageColumn``   ``image``            Appends 'Image Name' column
+Image Name   ``StringColumn``  ``s``                Appends 'Image' column
```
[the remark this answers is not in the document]
muhanadz: I had more details at first but rereading it sounded much more complex than it is. I don't think the user benefits from knowing the column types at all, if anything, this is good documentation for us.

README.rst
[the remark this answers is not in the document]
muhanadz: `ROI Name` currently gets auto-detected as 's' header type, thus `StringColumn`. It seems to fail to resolve even when you manually set the header type to `roi`:
csv:
```
image_id,Roi Name,size
322501,test,60.5
```
Resolved: 
image

README.rst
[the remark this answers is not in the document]
will-moore: OK. 👍

README.rst
```diff
@@ -64,45 +64,88 @@ populate
...
+Image ID     ``ImageColumn``   ``image``            Appends 'Image Name' column
+Dataset      ``DatasetColumn`` ``dataset``          \-
+Dataset Name ``StringColumn``  ``s``                \-
+Dataset ID   ``DatasetColumn`` ``dataset``          \-
+Plate        ``PlateColumn``   ``plate``            Adds 'Plate' column
```
will-moore, line 88: I think this should be `Adds a 'Plate Name' column` (from the example below)?

muhanadz: The `plate` column assumes the data is plate names, not plate ids. Since plate names are provided, `omero metadata` adds a new plate column for plate IDs. The confusion comes from the way it adds the column, which is replaced then appends whatever was replaced to the end of the table. Thus technically only added the plate ID column since the original was the plate name. Can attempt to reword the table to what type of column is added (ie names or ids)

sbesson: I am a bit torn between 1- it's important for ourselves to capture this behavior knowledge (as we want to fix it moving forward) and 2- users don't (and shoudn't) care :) 

If rewording is to tricky, I could imagine using the standard wording but adding a footnote

sbesson:
Reading the latest version of the README:

- the section starts with a thorough explanation of the  default behavior which is a combined of name-based detection + pandas-based detection. The addition of tables is very welcome and a great way to capture the different scenarios.
- then it defines the semantics of the `# header`  row allowing to override the default behavior by encoding the target column types in the CSV directly
- finally several examples are given. Interestingly, all examples  make use the `# header` row with a note mentioning that removing this row should result in the same outcome

At least to me, this reads like we are giving mixed messages about how users should structure their data.
How do people feel about revert the paradigm and remove the `# header` row from the CSV snippets? This means the examples would describe the outcome of the default behavior. In addition, each example could include a  note stating that the same behavior could be achieved by adding the custom `# header ...` row at the top of the CSV.

pwalczysko (commented):
Screenshot 2022-05-23 at 11 27 18
Change the `omero` in the line above to ``omero metadata``. This will fix the formatting and also will be more precise.

README.rst
```diff
@@ -64,45 +64,88 @@ populate
...
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-If you wish to ensure that ``number`` columns are created for numerical data, this will
-allow you to make numerical queries on the table.
-Column Types are:
+OMERO.tables have defined column types to specify the data-type such as ``double`` or ``long`` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
+
+The default behaviour of the script is to automatically detect the column types from an input ``CSV``. This behaviour works as follows:
+
+*  Columns named with a supported object-type (e.g. ``plate``, ``well``, ``image``, ``dataset``, or ``roi``), with ``<object>_id`` or ``<object>_name`` will generate the corresponding column type in the OMERO.table. See table below for full list of supported column names.
```
pwalczysko, line 77: ``<object>_id`` has an underscore in it. But in the table below, the column names have no underscores. How to reconcile this ?

pwalczysko: I can see that this is explained under the table, but it could give a much nicer read if the example and the table are unified also above and in the table, especially as it does not matter whether there is a space or underscore...

pwalczysko (changes requested):

pwalczysko:
> How do people feel about revert the paradigm and remove the # header row from the CSV snippets? 

The downside of this is that then we will need examples showing how the ``# header`` should look like. Unfortunately, this is not trivial and the users kept asking about it. I think a very simple example listed on the top could show the behavour with and without ``# header``, elucidating the (rare?) situations where the ``# header`` might be wanted/needed, rather than summarily removing all the ``# header`` examples ?

will-moore:
I think we should go with @sbesson suggestion: "each example could include a note stating that the same behavior could be achieved by adding the custom # header ... row at the top of the CSV". Then we don't lose that info.
I'm not sure we have any use cases where we think the header is needed, except "you're not using the latest version of the `omero-metadata`" which is a possibility. Maybe that should be mentioned in the README?

We need to update the `Populate_Metadata.py` script, to take advantage of the auto header detection and/or warn if not using a version that supports it.

sbesson:
> I'm not sure we have any use cases where we think the header is needed, except "you're not using the latest version of the omero-metadata" which is a possibility.

It is certainly my hope that @muhanadz's work is making the manual column annotation unnecessary in the majority of the cases. But I would not discard special cases where the column types need to be specified manually.
Once this PR is approved and merged, my vote is to get a `0.11.0` release out in order for both ourselves and the community to give the new functionality a try and provide feedback.

muhanadz:
I followed @sbesson 's suggestion by reverting the paradigm for the examples. I do believe this read better but happy to revert the changes if necessary.  

We will probably have to revisit the README and change the wording when more changes are made on the paths of the code.

sbesson (approved):
Thanks @muhanadz. The latest version reads well to me and is a substantial improvement in terms of explaining how the columns are mapped into OMERO.tables. My inclination is to get this merged, release the library/plugin and iterate both on the code and the README/documentation.

pwalczysko (changes requested):

pwalczysko:
Couple of suggestions, but looks good otherwise.

README.rst
```diff
@@ -114,23 +160,52 @@ img-03.png dataset01    0.093    3             TRITC        36640
...
+
+    image id,Dataset ID,ROI_Area,Channel_Index,Channel_Name
+    36638,101,0.0469,1,DAPI
+    36639,101,0.142,2,GFP
+    36640,101,0.093,3,TRITC
+    36641,101,0.429,4,Cy5
+
+
+The previous example will create an OMERO.table linked to the Project as follows with
+a new ``Image`` column with Names:
```
will-moore, line 177: The new column in the table below is ``Image Name``

will-moore:
Final read-through, looking good - couple of comments.
Also just noticed Copyright: "2018-2021, The Open Microscopy Environment" could be updated.

README.rst
```diff
@@ -64,45 +64,91 @@ populate
...
 - ``l``: ``LongColumn``, for integer numbers
 - ``s``: ``StringColumn``, for text
 - ``b``: ``BoolColumn``, for true/false
 - ``plate``, ``well``, ``image``, ``dataset``, ``roi`` to specify objects
 
-These can be specified in the first row of a ``CSV`` with a ``# header`` tag (see examples below).
-The ``# header`` row is optional. Default column type is ``String``.
+Automatic header detection can also be ignored if using the ``--manual_headers`` flag. If the ``# header`` is not present and this flag is used, column types will default to ``String`` (unless the column names correspond to OMERO objects such as ``image`` or ``plate``).
 
 NB: Column names should not contain spaces if you want to be able to query
```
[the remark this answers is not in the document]
muhanadz: Agreed. Addressed in https://github.com/ome/omero-metadata/pull/71/commits/b36691eacad2463b813b017d551259cc99392458

README.rst
```diff
@@ -158,17 +236,19 @@ NB: Columns of type ``shape`` aren't yet supported on the OMERO.server.
 
 Alternatively, if the target is an Image, the ROI input column can be
 ``Roi Name`` (with type ``s``), and an ``roi`` type column will be appended containing ROI IDs.
-In this case, it is required that ROIs on the Image in OMERO have the ``Name`` attribute set.
+In this case, it is required that ROIs on the Image in OMERO have the ``Name`` attribute set::
```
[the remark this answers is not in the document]
muhanadz: Agreed. Addressed in https://github.com/ome/omero-metadata/pull/71/commits/29a81030ddabbf628c9e2bcdefb4b5d80d47ea46

README.rst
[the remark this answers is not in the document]
muhanadz: Thanks, @will-moore. I found out that my validation wasn't done correctly there since OMERO.iviewer couldn't add ROI names but only comments. Updated table in [f8bcffd](https://github.com/ome/omero-metadata/pull/71/commits/f8bcffdb02e62f1a127193824f95b866a9178d99)

README.rst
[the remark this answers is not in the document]
muhanadz: Thank you for spotting that. Addressed in https://github.com/ome/omero-metadata/commit/08bd5d842cb5c027aa821f88cd306e9ee54b878c

README.rst
```diff
@@ -114,23 +160,52 @@ img-03.png dataset01    0.093    3             TRITC        36640
...
+===== ======= ======== ============= ============ ==========
+Image Dataset ROI_Area Channel_Index Channel_Name Image Name
+===== ======= ======== ============= ============ ==========
+36638 101     0.0469   1             DAPI         img-01.png 
+36639 101     0.142    2             GFP          img-02.png 
+36640 101     0.093    3             TRITC        img-03.png 
+36641 101     0.429    4             Cy5          img-04.png
+===== ======= ======== ============= ============ ==========
+
 If the target is a Dataset instead of a Project, the ``Dataset Name`` column is not needed.
```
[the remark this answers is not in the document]
muhanadz: Moved to after the `Note:...` and generalized the sentence to include both previous example in https://github.com/ome/omero-metadata/pull/71/commits/12cac5b5f3b560443099d2b9a94dc8e5cffdf323

README.rst
muhanadz: @sbesson @will-moore Coming back to this, I would add a footnote but there are already multiple notes under the table.
After much brainstorming, I gave it a good attempt in adding more information to the table (in https://github.com/ome/omero-metadata/pull/71/commits/04eda4ef665681a3ee54b03bceed65d3291505af) while capturing knowledge of the current behavior, staying minimalistic, and hopefully improving clarity for the reader/user. Feedback welcome on this change (as I've been staring at it for too long) and happy to revert if you think this is worse or more confusing.

muhanadz:
Added final changes from comments and suggestions (thank you all). Made some changes to the object-types table to address the conversation here: https://github.com/ome/omero-metadata/pull/71#discussion_r887379390

will-moore (approved):
All looks good now, thanks!

sbesson (approved):
Also happy for this to go in and I'll leave @pwalczysko to submit his review.

Once this is in, #77 is the only PR I would propose for review and possible inclusion in `0.11.0` as it fixes a specific (but not uncommon) use case related to the new detection feature. If there is no reviewing capacity, this can be scheduled for a patch release.

README.rst
```diff
@@ -64,45 +64,90 @@ populate
...
+the table as a ``File Annotation`` to a parent container such as Screen, Plate, Project,
 Dataset or Image. It also attempts to convert Image, Well or ROI names from the ``CSV`` into
 object IDs in the ``OMERO.table``.
 
 The ``CSV`` file must be provided as local file with ``--file path/to/file.csv``.
 
-If you wish to ensure that ``number`` columns are created for numerical data, this will
-allow you to make numerical queries on the table.
-Column Types are:
+OMERO.tables have defined column types to specify the data-type such as ``double`` or ``long`` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``
```
pwalczysko, line 73:
```suggestion
OMERO.tables have defined column types to specify the data-type such as ``double`` or ``long`` and special object-types of each column for storing OMERO object IDs such as ``ImageColumn`` or ``WellColumn``.
```

pwalczysko:
Sorry for the delay. Like the text, just one missing dot https://github.com/ome/omero-metadata/pull/71/files#r902463366 spotted. After accepting the suggestion https://github.com/ome/omero-metadata/pull/71/files#r902463366 all looks good to me here.

pwalczysko (changes requested):
Fix https://github.com/ome/omero-metadata/pull/71/files#r902463366 please, otherwise good to go.

pwalczysko (approved):

sbesson:
Thanks all for the time spent on this readme. Lots of back n forth but i think in addition to introducing the new detection, the readme now clarifies several behaviors of the plugin. 

From my side the only blocker to releasing 0.11.0 is the scenario of csv files with sparse numerical data which is currently broken when using the default command (see #76). #77 proposes a fix for this regression by adding a new parameter in the HeaderDetection API. @will-moore @muhanadz could you take a look at this and see if the proposal makes sense?

PR #75, opened 2022-05-17
Added column type tests

muhanadz:
While testing the README changes in https://github.com/ome/omero-metadata/pull/71, written some tests to assert the resolved column types from `omero metadata populate`. 

These tests provide insight and could help us align the expected behavior versus the executed behavior.

### DIFF
--- a/test/unit/test_automatic_header.py
+++ b/test/unit/test_automatic_header.py
@@ -7,16 +7,20 @@
 # file you can find at the root of the distribution bundle.  If the file is
 # missing please request a copy by contacting info@glencoesoftware.com
 
-"""
-   Test of the default automatic column type detection behaviour
-"""
 
+from omero.model import ScreenI, ProjectI
+from omero_metadata.populate import HeaderResolver
 from omero_metadata.cli import MetadataControl
 import pandas as pd
 import tempfile
+from omero.grid import ImageColumn, LongColumn, PlateColumn, RoiColumn, \
+    StringColumn, WellColumn, DoubleColumn, BoolColumn, DatasetColumn
 
 
 def test_detect_headers():
+    '''
+    Test of the default automatic column type detection behaviour
+    '''
     d = {
         'measurement 1': [11, 22, 33],
         'measurement 2': [0.1, 0.2, 0.3],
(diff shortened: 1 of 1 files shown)


sbesson:
My understanding is that both either the pandas based detection or the header row parsing will first a `column_types` list. This variable is then passed as an argument the `HeaderResolver` and resolved into OMERO.table columns. The goal of the class introduced in this PR is to add coverage for the last part of the logic.

@will-moore should https://github.com/ome/omero-metadata/pull/75#issuecomment-1129071045 not be a completely separate unit test, possibly re-using the logic of the tests introduced in https://github.com/ome/omero-metadata/pull/72 to test `#header` -> `column_types`?

test/unit/test_automatic_header.py
```diff
@@ -50,3 +54,174 @@ def test_detect_headers():
...
+        expected_resolved_column_names = [
+            'Plate', 'Well', 'measurement 1', 'measurement 2', 'measurement 3',
+            'measurement 4', 'Plate Name', 'Well Name']
+
+        expected_resolved_column_type = [
+            PlateColumn, WellColumn,
+            LongColumn, DoubleColumn, StringColumn, BoolColumn,
+            StringColumn, StringColumn]
+
+        target_object = ScreenI(0, None)  # Target object is Screen
```
sbesson, line 91: Based on my reading of the `HeaderResolver`, it looks like the  column type resolution is agnostic of the target object class if `column_types` is specified.

In other terms, the same test as this one with `target_object=DatasetI(0, None)` should have the same outcome (and the population logic fails further down the line when trying to resolve individual values).

Is that also your understanding and if so, is it worth capturing in these tests as well to cover the current behavior ?

will-moore:
@sbesson Yes, I guess so. I just seemed that we've already got a good set of column names and corresponding column types in these tests that could be re-used for testing auto-header. But happy for this to happen however it's easiest.

test/unit/test_automatic_header.py
muhanadz: That is my understanding as well. Mainly kept changing the `target_object` across tests to document the expected behavior and the targeted object in this case.

@sbesson Can you clarify what would you like to capture with these tests?

sbesson: For instance, do you foresee value in testing the columns from `test_plate_name_well_name` against other target objects than `ScreenI` e.g. `PlateI`, `DatasetI`...?

If not, restricting the unit tests to working scenarios described in the README as done in this PR is definitely the most important improvement and happy for this to get merged in its current form.

muhanadz: @sbesson Perhaps there could be the value I did not see. However, I only thought that as those tests will pass (because there is no check until further down the line as you mentioned) since that method is agnostic, it might be confusing.

We will have to rework those tests in the future if/when the code paths get reworked.

muhanadz: I updated the comment for `target_object` to be more accurate and descriptive of the behavior in https://github.com/ome/omero-metadata/pull/75/commits/d74b318d710c7ed277679901c53c2ec00f9153fd

sbesson (approved):
Merging this as a first step of documenting the behavior and notably the exceptation in terms of the mapping from CSV columns into OMERO.table columns (including added columns as described in #71).

As mentioned above, unit tests will need to be expanded as we add or fix the library behavior.

PR #77, opened 2022-06-05
Fix header detection for tables with sparse numerical data

### LINKED ISSUE #76: Header detection: default behavior and handling sparse columns
Discovered while testing the HEAD of the `omero-metadata` including the new header detection feature introduced in #67 against an IDR high-content screening dataset.

The annotation CSV contains a combination of biomolecular annotations (Organism, compound name, identifiers) and analytical metadata (features). The feature columns are densely populated but some of the biomolecular annotations are sparse e.g. `Compound Concentration (microMolar)`. This is expected since several rows correspond to control wells where there is no compound and this metadata is irrelevant.

The current version of the header detection code leads to issue in this case as these columns are detected as `Double/Float` and the table population subsequently fail unless `--allow-nan` is passed. With the current code, the workaround for completing the table population are:

- either to disable the manual header detection with `--manual_header`
- and/or to manually specify the behavior of the columns using the `#header` row

Ideally, it would be great to allow the plugin to "do the right thing" and handle these scenarios while retaining the automatic header detection to map the column the most appropriate type. This raises the question of whether there should be a single behavior or whether this would be another option down to the user.

In the IDR use case above, the expectation is that we want to preserve the sparsity rather than populating NaN values. Some downstream processes like the tables -> key/value conversion have logic depending on the emptiness of the values in the table so I expect NaN would cause issues (even though it reduces the table filtering ability).
I can conceive other use case where the user would like empty values to be interpreted as `NaN`.

Code-wise, it should be possible to use [pandas.read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) `keep_default_na` option to map such column as `object/StringColumn` rather than `float`:

```
(base) sbesson@Sebastiens-MacBook-Pro /tmp % cat test.csv 
Column1,Column2,Column3
A,1,2
B,,3
C,2,5%                                                                          
(base) sbesson@Sebastiens-MacBook-Pro /tmp % venv/bin/python
Python 3.8.11 (default, Jul 29 2021, 14:57:32) 
[Clang 12.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas
>>> df=pandas.read_csv('test.csv')
>>> df
  Column1  Column2  Column3
0       A      1.0        2
1       B      NaN        3
2       C      2.0        5
>>> df.dtypes
Column1     object
Column2    float64
Column3      int64
dtype: object
>>> df=pandas.read_csv('test.csv',keep_default_na=False)
>>> df
  Column1 Column2  Column3
0       A       1        2
1       B                3
2       C       2        5
>>> df.dtypes
Column1    object
Column2    object
Column3     int64
dtype: object
```

My current consideration would be to re-use the existing `--allow-nan` flag and set `keep_default_na` accordingly. 
@muhanadz @pwalczysko @will-moore happy to open a PR with accompanying unit tests. Any opinions on this?

sbesson:
Fixes #76 

## Reproducible scenario

First create a minimal dataset/image hierarchy e.g. as follows:

```bash
touch test1.fake test2.fake
dataset=$(omero obj new Dataset name=sparse_table)
omero import -T $dataset test1.fake test2.fake
```

CSV files with sparse string data such as the one below are correctly handled by the current HEAD of `omero-metadata`.

```
$ cat sparse_string_column.csv 
Image name,meas1,meas2,meas3,meas4
test1.fake,1.1,1,high,low
test2.fake,0.5,2,,low
```

The columns with missing values are mapped as `s/StringCOlumn` and the missing value are turned into empty strings where running `omero metadata populate --file sparse_string_column.csv $dataset` e.g.

CSV files with sparse numerical columns such as the one below currently fail during the population command:

```csv
$ cat sparse_numeric_column.csv
Image name,meas1,meas2,meas3,meas4
test1.fake,1.1,1.2,high,low
test2.fake,,2.1,,low
```

Here, the `meas1` column is currently mapped into a `d` header type/`DoubleColumn` by the `pandas` detection logic, With the default `omero metadata populate --file command, the table population fails with `ValueError: Empty Double or Long value. Use --allow_nan to convert to NaN`.

## Proposed changes

Since the library already includes some logic allowing the user to control whether NaN values are allowed in the OMERO.table (introduced in #60), this PR proposes the following changes

- the `MetadataControl.detect_headers` API supports an extra `keep_default_na` argument (`True` by default)`. Its value is forwarded to `pandas.read_csv` and determines controls how pandas should handle NA (missing) values and whether the column is detected as `d` vs `s` (see https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html)
- unit tests are added to cover the new `MetadataControl.detect_headers` API with different combination of tables/arguments
- the populate command is updated to use the value of `args.allow_nan` (`False` by default) as pass it as  `keep_default_na` to `MetadataControl.detect_headers` 

fe73a17d2a71fd7d220c48891a2364110b59f4f1 adds a cosmetic change defining GNU-style aliases of the command-line arguments (`--manual-header`, `--allow-nan`) using hyphen as separator. The existing underscore separated flags are preserved.

## Testing

With these changes, annotating of sparse CSV tables using the default header detection should be functional in all cases.

1. if the tabular data is dense or containing sparse string columns, the behavior of the command should be unchanged 
2. if the tabular data is dense or containing sparse numerical columns,, the behavior of the command  will depend on the  `--allow-nan` flag

   ```
   omero metadata populate --file sparse_numeric_column.csv $data
   ```
   will detect the sparse numeric column as a`StringColumn` and store the missing values as empty strings

   ```
   omero metadata populate --file sparse_numeric_column.csv --allow-nan $data
   ```
   will detect the sparse numeric column as a `DoubleColumn` and store missing values as `nan`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -111,6 +111,9 @@ Example Float   ``DoubleColumn``  ``d``
 Example boolean ``BoolColumn``    ``b``      
 =============== ================= ====================
 
+In the case of missing values, the column will be detected as ``StringColumn`` by default. If ``--allow-nan`` is passed to the
+``omero metadata populate`` commands, missing values in floating-point columns will be detected as ``DoubleColumn`` and the
+missing values will be stored as NaN.
 
 However, it is possible to manually define the header types, ignoring the automatic header detection, if a ``CSV`` with a ``# header`` row is passed. The ``# header`` row should be the first row of the CSV and defines columns according to the following list (see examples below):
 
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -241,11 +241,13 @@ class MetadataControl(BaseControl):
         populate.add_argument("--localcfg", help=(
             "Local configuration file or a JSON object string"))
 
-        populate.add_argument("--allow_nan", action="store_true", help=(
-            "Allow empty values to become Nan in Long or Double columns"))
+        populate.add_argument(
+           "--allow-nan", "--allow_nan", action="store_true", help=(
+                "Allow empty values to become Nan in Long or Double columns"))
 
-        populate.add_argument("--manual_header", action="store_true", help=(
-            "Disable automatic header detection during population"))
+        populate.add_argument(
+            "--manual-header", "--manual_header", action="store_true", help=(
+                "Disable automatic header detection during population"))
 
         populateroi.add_argument(
             "--measurement", type=int, default=None,
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -489,7 +489,7 @@ class MetadataControl(BaseControl):
             self.ctx.die(100, "Failed to initialize Table")
 
     @staticmethod
-    def detect_headers(csv_path):
+    def detect_headers(csv_path, keep_default_na=True):
         '''
         Function to automatically detect headers from a CSV file. This function
         loads the table to pandas to detects the column type and match headers
@@ -497,7 +497,7 @@ class MetadataControl(BaseControl):
 
         conserved_headers = ['well', 'plate', 'image', 'dataset', 'roi']
         headers = []
-        table = pd.read_csv(csv_path)
+        table = pd.read_csv(csv_path, keep_default_na=keep_default_na)
         col_types = table.dtypes.values.tolist()
         cols = list(table.columns)
 
--- a/src/omero_metadata/cli.py
+++ b/src/omero_metadata/cli.py
@@ -579,7 +579,8 @@ class MetadataControl(BaseControl):
         if not args.manual_header and \
                 not first_row[0].str.contains('# header').bool():
             omero_metadata.populate.log.info("Detecting header types")
-            header_type = MetadataControl.detect_headers(args.file)
+            header_type = MetadataControl.detect_headers(
+                args.file, keep_default_na=args.allow_nan)
             if args.dry_run:
                 omero_metadata.populate.log.info(f"Header Types:{header_type}")
         else:
--- a/test/integration/metadata/test_populate.py
+++ b/test/integration/metadata/test_populate.py
@@ -175,6 +175,13 @@ class Fixture(object):
         col_names = "Well,Well Type,Concentration,Well Name"
         assert col_names == ",".join([c.name for c in columns])
 
+    def assert_values(self, row_values):
+        # Unsure where the lower-casing is happening
+        if "A1" in row_values or "a1" in row_values:
+            assert "Control" in row_values
+        elif "A2" in row_values or "a2" in row_values:
+            assert "Treatment" in row_values
+
     def assert_child_annotations(self, oas):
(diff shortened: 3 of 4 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-plugins-push#1169](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/1169/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/1169/consoleText) for more details.
Possible conflicts:
  - PR #64 will-moore 'Support images_by_name for Screen and Plate'
    - test/integration/metadata/test_populate.py

--conflicts

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-plugins-push#1176](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/1176/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-plugins-push/1176/consoleText) for more details.
Possible conflicts:
  - PR #64 will-moore 'Support images_by_name for Screen and Plate'
    - test/integration/metadata/test_populate.py

--conflicts

muhanadz (commented):
## Review summary
In conclusion, what was described in the PR description is reasonable, valid, and in my opinion necessary.

Following the steps described, without the PR there are cases where the script will fail (with a `ValueError` exception) in the presence of a sparse numeric column and the lack of `--allow-nan`. This PR fixes that behavior such that in the case of a sparse numeric column you can choose to have the sparse numeric column read as `'d'` (`DoubleColumn`) or `'s'` (`StringColumn`) by passing `--allow-nan` or not

My current understanding is that nan values or empty values will be processed regardless of if they're in a numeric or string column. And the default is to read the nan values as string type.

If so I'd suggest adding something like "without `--allow-nan`, by default, the column will be read as a string column in the presence of nans in the column" somewhere either in the help of the argument or the README.

## Review details
### Using `sparse_string_column.csv` (4th column sparse):
Without PR detected `['s', 'd', 'l', 's', 's']`. Added empty string to final result.
With PR changes, detected `['s', 'd', 'l', 's', 's']` and with `--allow-nan`, similary detected `['s', 'd', 'l', 's', 's']`. Added empty string to final result.


### Using `sparse_numeric_column.csv (2nd and 4th column sparse)`
Without PR detected `['s', 'd', 'd', 's', 's']`, however will get the following exception:
```
ValueError: Empty Double or Long value. Use --allow_nan to convert to NaN
```
With PR changes, detected `['s', 's', 'd', 's', 's']` which a dded empty string to final result and with `--allow-nan` detected `['s', 'd', 'd', 's', 's']` which added `'nan'` to final result.

muhanadz (approved):
Looks good to me

sbesson:
Thanks @muhanadz. The new detection behavior is now releases as `omero-metadata 0.11.0` 🎉

PR #78, opened 2022-06-28
"Nothing to do during post processing." from info to debug

emilroz:
This particular line is creating as many entries in the log file as there is rows in the table! This is particularly bad in the case where number of rows in the tables reach >100k.

If there's no particular reason to keep it at the info level it would be great to change it to debug.

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -1401,7 +1401,7 @@ class ParsingContext(object):
         if well_name_column is None and plate_name_column is None \
                 and image_name_column is None and roi_name_column is None \
                 and roi_column is None:
-            log.info('Nothing to do during post processing.')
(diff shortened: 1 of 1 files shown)


sbesson (approved):
I completely agree this is DEBUG level especially given the scale at which this utility is executed these days but I will leave @chris-allan who introduced the original logic https://github.com/ome/omero-metadata/commit/94db760ece176aa494c75228d7a44c4694cca58d a chance to confirm. 

Once approved, proposed to get this out as an immediate patch release?

chris-allan (approved):
Once upon a time it made very little sense to not have one of these "major" columns. That time has long since passed.

sbesson:
Now released as https://pypi.org/project/omero-metadata/0.11.1/

PR #79, opened 2022-09-02
Remove leading and trailing whitespaces from csv fields and headers

dominikl:
@francesw and @pwalczysko noticed that we have annotations with leading/trailing whitespaces in IDR. Is there any reason why it should be possible to have entries with leading/trailing whitespaces in the first place? If not it would be best to remove them directly in the metadata plugin.


joshmoore:
and/or is it fair to say that _if_ leading/trailing whitespace is needed, then quotes should be used?

dominikl:
Not sure how to implement that. Can't tell if quotes were used or not, because they're removed by the csv library.

joshmoore:
Aha. Then perhaps we delegate more of the work to `csv`? https://docs.python.org/2/library/csv.html#csv.Dialect.skipinitialspace

dominikl:
Thought about that too. But it looks like this only deals with leading space, doesn't it? But that might be sufficient?

joshmoore:
I'm sure we'll eventually have trailing spaces (if we don't already...). Guess we're back to your question if it's _ever_ valid to want leading space, because if so and the quotes have already been parsed away, then with this trimming, there will be no way for someone to preserve whitespace.

dominikl:
Reverted the commit, and added the `skipinitialspace` option. Like expected, leading spaces get removed but trailing spaces don't, spaces in quotes get preserved. 

```
rows |  
-- | --
0 |  
0 | "metadata_test"
1 | "test-1.fake"
2 | " with quotes "
3 | 238309
1 |  
0 | "metadata_test"
1 | "test-2.fake"
2 | "without quotes"
3 | 238310
2 |  
0 | "metadata_test"
1 | "test-3.fake"
2 | "with trailing space "
3 | 238311
```

joshmoore:
Do we know of instances of trailing space that we need to deal with? and/or are the problem cases we're currently struggling with quoted?

pwalczysko:
> Do we know of instances of trailing space that we need to deal with? and/or are the problem cases we're currently struggling with quoted?

No, the cases with which this was triggered were not quoted, see screenshots


Screenshot 2022-09-13 at 14 51 47
Screenshot 2022-09-13 at 14 51 18

pwalczysko (approved):

pwalczysko:
> My understanding of the decision is to implement the following behavior:
> 
>     * if a CSV contains unquoted strings with leading spaces in any row (header or not), these spaces will be removed during by the tables population step
> 
>     * if a CSV contains quoted strings with leading spaces, they will be preserved

This matches my experience when testing this PR.

test/integration/metadata/bulk_to_map_annotation_context_ns.csv
```diff
@@ -6,4 +6,8 @@ a4,FBgn0086378,,Alg-2
 b1,,hh,DHH;IHH;SHH;Desert hedgehog;Indian hedgehog;Sonic hedgehog
 b2,,sws,PNPLA6;patatin like phospholipase domain containing 6
 b3,,ken,BCL6;B-cell lymphoma 6 protein
-b4,,,Alg-2
\ No newline at end of file
+b4,,,Alg-2
+c1,a, leading whitespace,one
+c2,b,trailing whitespace ,two
+c3,c, leading and trailing whitespaces ,three
+c4,d," leading and trailing whitespaces in quotes ",four
```
jburel, line 13: Can you an extra line at the end of this file

dominikl: Reverted the change again, doesn't seem to work that way anyway.

will-moore:
Looks like 3 tests are failing, all with:

```
    def assert_columns(self, columns):
        col_names = "Well,Well Type,Concentration,Well Name"
>       assert col_names == ",".join([c.name for c in columns])
E       AssertionError: assert 'Well,Well Ty...ion,Well Name' == 'Well,FlyBase,...mes,Well Name'
E         - Well,Well Type,Concentration,Well Name
E         + Well,FlyBase,Gene,Gene Names,Well Name
```

test/integration/metadata/test_populate.py
```diff
@@ -669,6 +669,35 @@ def get_namespaces(self):
...
+    def assert_row_values(self, rowvalues):
+        assert rowvalues[0][1:] == (
+            "a1", "a", "leading whitespace", "one")
+        assert rowvalues[1][1:] == (
+            "a2", "b", "trailing whitespace ", "two")
+        assert rowvalues[2][1:] == (
+            "a3", "c", "leading and trailing whitespaces ", "three")
+        assert rowvalues[3][1:] == (
+            "a4", "d", " leading and trailing whitespaces in quotes ", "four")
+
```
will-moore, line 699: I think you need to add this to your class

```
def assert_columns(self, columns):
        col_names = "Well,FlyBase,Gene,Gene Names"
        assert col_names == ",".join([c.name for c in columns])
```

dominikl: Ah thanks Will, was wondering why it was still failing 👍

will-moore:
Nearly there!

```
>       assert col_names == ",".join([c.name for c in columns])
E       AssertionError: assert 'Well,FlyBase,Gene,Gene Names' == 'Well,FlyBase,...mes,Well Name'
E         - Well,FlyBase,Gene,Gene Names
E         + Well,FlyBase,Gene,Gene Names,Well Name
```

dominikl:
Weird, wonder where the `Well Name` comes from... but I can add it.

dominikl:
That's getting silly. I'll remove the "test" commits again. Can be added in a separate PR once I found out how to add a simple integration test to test_populate.py.

will-moore:
Discussed in idr meeting today - found cases of trailing white-spaces, `"4,2,4'-Tribenzyloxy-6'-hydroxychalcone "` that are breaking search.
These should be removed along with leading whitespaces.

will-moore:
Aim to fix in IDR and release by end of October.

sbesson:
Note this PR aims at patching our annotation tools to ensure _new_ tables do not include trailing whitespaces. This is related but separate from the effort discussed above which aims at ensuring that existing tables do not have trailing whitespaces and re-create them accordingly.

Do we aim to have this fix included in a patch release of `omero-metadata` and deployed on the new IDR systems around the same timeline?

will-moore:
@dominikl I'm not sure where you got to on adding tests, but it seems you could simply add a single white-space to one of the existing tests, and it should have no effect on the created tables?

PR #82, opened 2023-04-17
Update ParsingContext.parse() to return the ID of the table

sbesson:
Comes from a discussion with @muhanadz. For scenarios where the `omero-metadata` plugin is used as a Python library, it is currently a requirement to make an additional query to retrieve the table created by `omero_metadata.populate.ParsingContext.parse()` operation.

As the implementation passes through the return value of `parse_from_handle_stream` https://github.com/ome/omero-metadata/blob/0ea482289812abf6adc58e13d2a49f71874d51d9/src/omero_metadata/populate.py#L1208-L1218, this PR amends the latter API to return the ID of the original file associated with the OMERO.table.

The unit tests are also amended to compare the output of the `parse()` API with the query and give an example of how a consumer might use the new functionality.

### DIFF
--- a/src/omero_metadata/populate.py
+++ b/src/omero_metadata/populate.py
@@ -1184,12 +1184,12 @@ class ParsingContext(object):
             self.populate_from_reader(reader,
                                       self.filter_function, table, 1000)
             self.create_file_annotation(table)
+            log.debug('Column widths: %r' % self.get_column_widths())
+            log.debug('Columns: %r' % [
+                (o.name, len(o.values)) for o in self.columns])
+            return table.getOriginalFile().id.val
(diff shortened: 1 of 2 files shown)


muhanadz (approved):
Works as expected. The update to `ParsingContext.parse()` to return the ID of the table is a practical and useful improvement for scenarios where the `omero-metadata` plugin is used as a Python library which happens often (example: https://github.com/ome/omero-scripts/blob/develop/omero/import_scripts/Populate_Metadata.py#L132). The changes to the unit tests provides a clear example of how to leverage the new returned value.

jburel:
Please do go ahead with a release

PR #83, opened 2023-11-14
Remove mox3 test dependency

sbesson:
Follow-up of https://github.com/ome/omero-py/pull/385, this removes the (unused) `mox3` dependency from the test requirements of this plugin.

Note that the current OMERO.server Docker image used by the https://github.com/ome/omero-test-infra infrastructure ships an outdated version of OMERO.py (as in older than the just released 5.17.0). f6a286e7063aedf744ffbf013c7760fabe8fe335 overrides `py-setup` file to upgrade OMERO.py and include the `testlib` changes deprecating the `mox3` usage.  If the decision during the review process is to re-build these images, the last commit should be unnecessary and can be force pushed away.


jburel:
test-infra also creates issues due to the version of Python installed

sbesson:
[3383fe1](https://github.com/ome/omero-metadata/pull/83/commits/3383fe1b96af00ef3992a07080f9d1dcaa14fb7b) is probably a more elegant alternative to overriding `py-setup` as it declares the minimal OMERO.py version required for the tests and allows to remove the `mox3` dependency entirely.  At least on my fork, all tests were passing https://github.com/sbesson/omero-metadata/actions/runs/6888372479/job/18737236012.

Note that GitHub workflows have been deactivated and will need to be re-enabled on this repository (as well as `omero-cli-render`, `omero-upload` and `omero-cli-duplicate` where similar changes will need to be ported.

PR #84, opened 2024-01-11
Fix flake8 issue raised following the Rocky Linux 9 image upgrade

sbesson:
See https://github.com/ome/omero-metadata/actions/runs/7490016877/job/20387910739

I assume the new operating system is pulling a more recent of `flake8` with additional warnings

PR #85, opened 2024-01-22
Integration tests: call pytest directly

sbesson:
See https://github.com/ome/omero-metadata/pull/84#issuecomment-1888767870

Following the upgrade of the test infrastructure to Rocky Linux 9/Python 3.9 (see https://github.com/ome/omero-test-infra/pull/66), some of the tests started failing due to the presence of logging statements in the stdout conflicting with some of the `assert` checks.
After some investigation, this behavioral change does not look to be related to the operating system or Python version per se but is more specific to the testing infrastructure.

These changes depend on https://github.com/ome/omero-test-infra/pull/68 to call `pytest` directly instead of `python setup.py test` and revert a11343c71bf89d4d116137e1db96dff7e54e51ee.

### DIFF
--- a/test/integration/clitest/test_metadata.py
+++ b/test/integration/clitest/test_metadata.py
@@ -20,7 +20,6 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 import pytest
-from past.builtins import long
 
 import omero
 import omero.gateway
@@ -43,7 +42,7 @@ class MetadataTestBase(CLITest):
 
         conn = BlitzGateway(client_obj=self.client)
         self.imageid = unwrap(self.image.getId())
-        assert isinstance(self.imageid, long)
+        assert isinstance(self.imageid, int)
         wrapper = conn.getObject("Image", self.imageid)
         self.md = Metadata(wrapper)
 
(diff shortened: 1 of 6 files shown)


test/integration/clitest/test_metadata.py
```diff
@@ -245,9 +244,10 @@ def test_deletebulkanns(self, capfd):
         o = self.invoke(capfd)
         assert "FileAnnotation:" in o
 
-        # Should have no FileAnnotation since it's deleted
+        # Should be empty since it's deleted
         o = self.invoke(capfd)
-        assert "FileAnnotation:" not in o
+        assert o == ""
+        assert len(o.strip()) == 0
```
will-moore, line 250: This 2nd `assert` seems redundant since you're already checked that `o == ""`?

will-moore (approved):
Code changes look good and integration tests with omero-test-infra are passing.

test/integration/clitest/test_metadata.py
sbesson: The tests have been restored to their previous state before cd4f65741c105afb8aaa839c2a631777f60bf585 but I agree the second assert feels likely superfluous and could be pushed away if we wish so

sbesson:
A side-note that these changes also include the removal of the `future` dependency in [a20206d](https://github.com/ome/omero-metadata/pull/85/commits/a20206dff85b5c802a05b61df1f435f80e91e23e).

Combined with the [release of OMERO.py 5.19.0](https://github.com/ome/omero-py/blob/v5.19.0/CHANGELOG.md), this should make this plugin functional with Python 3.12.0. Adding it to the overarching GitHub projects tracking the various locations to update/review

PR #86, opened 2024-08-29
Add tablename option

dominikl:
Allow to specify a custom table name for `metadata populate` command.


will-moore:
Looks good:

```
$ omero metadata populate Project:7707 --file project_7707_mergeci.csv --table-name wills_data2
```
Also checked that without the `--table-name` you get the default name:

Screenshot 2024-08-30 at 10 47 31


I was going to suggest that you add a test for this, but I don't see any tests that invoke the command line, and the `table_name` parameter is already tested.

will-moore (approved):
LGTM 👍

PR #89, opened 2024-11-20
Build

jburel:
Update actions version and Update PyPI publish workflow to use build instead of sdist


dominikl (approved):
👍 Looks good to me!

